PR #5, opened 2020-01-20
Discovery integration

shahankhatch:
Enable findClosestNodes to dump the entire discovered peer table


src/test/java/org/ethereum/beacon/discovery/NodeRecordTest.java
```diff
@@ -54,7 +53,7 @@ public void testLocalhostV4() throws Exception {
         InetAddress.getByName(expectedHost).getAddress(),
         ((Bytes) nodeRecord.get(EnrField.IP_V4)).toArray());
     assertEquals(expectedUdpPort, nodeRecord.get(EnrField.UDP_V4));
-    assertEquals(expectedTcpPort, nodeRecord.get(EnrField.TCP_V4));
+    //    assertEquals(expectedTcpPort, nodeRecord.get(EnrField.TCP_V4));
```
ajsutton: What caused these lines in the tests to be commented out? I can't see how it's related to the logLimit change.

shahankhatch: Unrelated to this PR, I've removed the TCP-related ports from UDP-based discovery services.

PR #12, opened 2020-02-18
ENR records should not include padding when serializing to base 64

ajsutton:
## PR Description
Omit padding characters when creating enr records as base64.  Add tests to ensure the ENR is serialised and parsed correctly based on the spec test vectors.

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
+++ b/src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
@@ -106,10 +106,8 @@ public class NodeRecord {
 
   public String asBase64() {
     String result = Base64.getUrlEncoder().encodeToString(serialize().toArray());
-    while (result.endsWith("=")) {
-      result = result.substring(0, result.length() - 1);
-    }
-    return result;
+    final int firstPaddingChar = result.indexOf('=');
+    return firstPaddingChar > 0 ? result.substring(0, firstPaddingChar) : result;
   }
 
   public String asEnr() {
--- a/src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
+++ b/src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
@@ -104,9 +104,16 @@ public class NodeRecord {
     return nodeRecord;
   }
 
-  @SuppressWarnings({"unchecked", "DefaultCharset"})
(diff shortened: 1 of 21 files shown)


src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
```diff
@@ -104,9 +104,16 @@ public static NodeRecord fromRawFields(
     return nodeRecord;
   }
 
-  @SuppressWarnings({"unchecked", "DefaultCharset"})
   public String asBase64() {
-    return new String(Base64.getUrlEncoder().encode(serialize().toArray()));
+    String result = Base64.getUrlEncoder().encodeToString(serialize().toArray());
+    while (result.endsWith("=")) {
```
mbaxter: Is it a bad idea to just search for the first index of "=" and truncate from there?

src/test/java/org/ethereum/beacon/discovery/NodeRecordTest.java
```diff
@@ -114,4 +115,50 @@ public void testSignature() throws Exception {
...
+                UInt64.valueOf(seq),
+                Pair.with(EnrField.ID, IdentitySchema.V4),
+                Pair.with(EnrFieldV4.IP_V4, ip),
+                Pair.with(EnrFieldV4.UDP_V4, port),
+                Pair.with(
+                    EnrFieldV4.PKEY_SECP256K1, Functions.derivePublicKeyFromPrivate(privateKey)));
+    nodeRecord.sign(privateKey);
+    assertEquals(nodeId, nodeRecord.getNodeId());
+    final Bytes bytes = nodeRecord.serialize();
+    System.out.println(bytes);
```
mbaxter: Stray `println`

src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
ajsutton: No that makes sense. `=` will only ever appear as a padding char at the so that would be faster and simpler.

src/test/java/org/ethereum/beacon/discovery/NodeRecordTest.java
ajsutton: There's always one that sneaks through...

PR #17, opened 2020-02-24
Prepare for Integration

ajsutton:
## PR Description
A number of changes to make the API easier to use:

* Replace use of `Pair` for tuples library with dedicated classes.
* Introduce NodeRecordBuilder to encapsulate knowledge of the standard EnrFields
* Bundle DiscoveryManager and DiscoveryTaskManager into a single DiscoverySystem with a method to retrieve all known peers


src/main/java/org/ethereum/beacon/discovery/schema/EnrField.java
```diff
@@ -4,20 +4,65 @@
 
 package org.ethereum.beacon.discovery.schema;
 
+import com.google.common.base.MoreObjects;
+import java.util.Objects;
+
 /** Fields of Ethereum Node Record */
```
mbaxter: Might be worth leaving a pointer to the docs here: https://github.com/ethereum/devp2p/blob/master/enr.md

src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
```diff
@@ -63,9 +60,9 @@ private NodeRecord(IdentitySchemaInterpreter identitySchemaInterpreter, UInt64 s
   public static NodeRecord fromValues(
       IdentitySchemaInterpreter identitySchemaInterpreter,
       UInt64 seq,
-      List<Pair<String, Object>> fieldKeyPairs) {
+      List<EnrField> fieldKeyPairs) {
```
mbaxter: nice :D

PR #21, opened 2020-03-10
Ensure find nodes request completes when there are no results

ajsutton:
## PR Description
When a `FINDNODE` request was sent that didn't find any nodes, the returned `CompletableFuture` would time out instead of completing successfully.

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/message/handler/NodesHandler.java
+++ b/src/main/java/org/ethereum/beacon/discovery/message/handler/NodesHandler.java
@@ -33,7 +33,7 @@ public class NodesHandler implements MessageHandler<NodesMessage> {
         requestInfo.getRemainingNodes() == null
             ? message.getTotal() - 1
             : requestInfo.getRemainingNodes() - 1;
-    if (newNodesCount == 0) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/ethereum/beacon/discovery/message/handler/NodesHandler.java
```diff
@@ -33,7 +33,7 @@ public void handle(NodesMessage message, NodeSession session) {
         requestInfo.getRemainingNodes() == null
             ? message.getTotal() - 1
             : requestInfo.getRemainingNodes() - 1;
-    if (newNodesCount == 0) {
+    if (newNodesCount <= 0) {
```
mbaxter: Why would this be < 0 ?  Shouldn't the responding node always include itself in the response so the total is at least 1?

ajsutton: Discv5 only includes nodes *at* the specified distance. So you get the node's own record by itself if you ask for distance 0 and will never get it if you ask for any other distance.  It's weird and mostly quite unhelpful.

PR #22, opened 2020-03-12
Implement simple recursive lookup

ajsutton:
## PR Description
Periodically perform a recursive lookup to find other nodes.  This is a minimal implementation that selects the closest of all known nodes to query and doesn't partition the search and only query nodes returned by previous steps in the search.  This is enough to discover the network, partitioning the search would improve sybil resistance - will follow up in a future PR once the basics are fully working.

This is confirmed to work with the `rumor` network testing tool but doesn't currently work with other discovery nodes - the responses to find nodes requests don't appear to be correct yet. Fixing the responses is being left to a future PR too.

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/task/RecursiveLookupTask.java
+++ b/src/main/java/org/ethereum/beacon/discovery/task/RecursiveLookupTask.java
@@ -1,0 +1,84 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+package org.ethereum.beacon.discovery.task;
+
+import static com.google.common.base.Preconditions.checkArgument;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.CompletableFuture;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes;
+import org.ethereum.beacon.discovery.schema.NodeRecordInfo;
+import org.ethereum.beacon.discovery.storage.NodeTable;
+import org.ethereum.beacon.discovery.util.Functions;
+
+public class RecursiveLookupTask {
+  private static final Logger LOG = LogManager.getLogger();
+  private static final int MAX_CONCURRENT_QUERIES = 3;
+  private final NodeTable nodeTable;
+  private final FindNodesAction sendFindNodesRequest;
+  private final Bytes targetNodeId;
+  private final Set<Bytes> queriedNodeIds = new HashSet<>();
+  private int availableQuerySlots = MAX_CONCURRENT_QUERIES;
+  private final CompletableFuture<Void> future = new CompletableFuture<>();
+
+  public RecursiveLookupTask(
+      final NodeTable nodeTable,
+      final FindNodesAction sendFindNodesRequest,
+      final Bytes targetNodeId) {
+    this.nodeTable = nodeTable;
+    this.sendFindNodesRequest = sendFindNodesRequest;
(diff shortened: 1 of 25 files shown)


src/main/java/org/ethereum/beacon/discovery/task/RecursiveLookupTasks.java
```diff
@@ -33,36 +32,37 @@ public RecursiveLookupTasks(
...
     scheduler.execute(
         () -> {
-          CompletableFuture<Void> retry =
-              discoveryManager.findNodes(nodeRecordInfo.getNode(), DEFAULT_DISTANCE);
-          taskTimeouts.put(
-              nodeRecordInfo.getNode().getNodeId(),
-              () ->
-                  retry.completeExceptionally(
-                      new RuntimeException("Timeout for node recursive lookup task")));
+          CompletableFuture<Void> retry = discoveryManager.findNodes(nodeRecord, distance);
```
mbaxter: Am I reading this wrong?  This doesn't look like a retry:
```suggestion
          CompletableFuture<Void> findNodesFuture = discoveryManager.findNodes(nodeRecord, distance);
```

src/main/java/org/ethereum/beacon/discovery/task/RecursiveLookupTask.java
```diff
@@ -0,0 +1,89 @@
...
+    nodeTable
+        .streamClosestNodes(targetNodeId, 0)
+        .filter(DiscoveryTaskManager.RECURSIVE_LOOKUP_NODE_RULE)
+        .filter(record -> !queriedNodeIds.contains(record.getNode().getNodeId()))
+        .limit(Math.min(availableQuerySlots, remainingTotalQueryLimit))
+        .forEach(this::queryPeer);
+    if (availableQuerySlots == MAX_CONCURRENT_QUERIES) {
+      // We didn't send any new queries so must have run out of possible nodes to query or reached
+      // the query limit.
+      future.complete(null);
```
mbaxter: Should we wait and only complete when all of the queries complete?

src/main/java/org/ethereum/beacon/discovery/task/RecursiveLookupTasks.java
ajsutton: That is indeed not a retry but for some reason that's what it was called.  Somewhere in the changesets I've been wrangling I've renamed it but apparently it didn't make it to this PR. Updated here to make sure it doesn't get forgotten.

src/test/java/org/ethereum/beacon/discovery/task/RecursiveLookupTaskTest.java
```diff
@@ -0,0 +1,255 @@
...
+    verifyNoMoreInteractions(findNodesAction);
+    assertTrue(complete.isDone());
+
+    // Complete remaining requests
+    findNodeRequests.get(PEER2).complete(null);
+    findNodeRequests.get(PEER3).complete(null);
+
+    verifyNoMoreInteractions(findNodesAction);
+    // Should now be done because all nodes have been queried
+    assertTrue(complete.isDone());
```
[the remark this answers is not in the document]
ajsutton: Fixed.

src/main/java/org/ethereum/beacon/discovery/task/RecursiveLookupTask.java
ajsutton: When all available query slots are unused (ie available slots = max concurrent queries) then there can't be any outstanding requests to wait for.  Clarified the comment.

PR #23, opened 2020-03-12
Validate the distance of returned nodes matches what we requested

ajsutton:
## PR Description
To make it more difficult for malicious nodes to pollute our node table, verify that any nodes returned from a find nodes request are at the distance requested.

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/message/handler/NodesHandler.java
+++ b/src/main/java/org/ethereum/beacon/discovery/message/handler/NodesHandler.java
@@ -54,22 +54,11 @@ public class NodesHandler implements MessageHandler<NodesMessage> {
             String.format(
                 "Received %s node records in session %s. Total buckets expected: %s",
                 message.getNodeRecordsSize(), session, message.getTotal()));
-    message
-        .getNodeRecords()
+    message.getNodeRecords().stream()
+        .filter(this::isValid)
+        .filter(record -> hasCorrectDistance(session, requestInfo, record))
         .forEach(
             nodeRecordV5 -> {
-              if (!nodeRecordV5.isValid()) {
-                logger.debug("Rejecting invalid node record {}", nodeRecordV5);
-                return;
-              }
-              if (hasIncorrectDistance(session, requestInfo, nodeRecordV5)) {
-                logger.debug(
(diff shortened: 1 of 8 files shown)


src/main/java/org/ethereum/beacon/discovery/message/handler/NodesHandler.java
```diff
@@ -52,19 +54,42 @@ public void handle(NodesMessage message, NodeSession session) {
...
+  }
+
+  private boolean hasCorrectDistance(
+      final NodeSession session,
+      final FindNodeRequestInfo requestInfo,
+      final NodeRecord nodeRecordV5) {
+    final int actualDistance = Functions.logDistance(nodeRecordV5.getNodeId(), session.getNodeId());
+    final int requestedDistance = requestInfo.getDistance();
+    if (actualDistance != requestedDistance) {
+
```
mbaxter: (nit) stray newline

src/test/java/org/ethereum/beacon/discovery/message/handler/NodesHandlerTest.java
```diff
@@ -0,0 +1,94 @@
...
+    handler.handle(message, session);
+
+    verifyNoInteractions(nodeTable);
+  }
+
+  @Test
+  public void shouldRejectReceivedRecordsThatAreNotAtCorrectDistance() {
+    final NodeInfo nodeInfo = TestUtil.generateNode(9000);
+    final FindNodeRequestInfo requestInfo =
+        new FindNodeRequestInfo(TaskStatus.SENT, REQUEST_ID, new CompletableFuture<>(), 0, null);
```
mbaxter: (optional) Distance of 0 feels like sort of a special value.  Might be worth explicitly calculating the `distance` and setting this to:
```suggestion
        new FindNodeRequestInfo(TaskStatus.SENT, REQUEST_ID, new CompletableFuture<>(), distance - 1, null);
```

ajsutton: Zero is special - it specifically means a request for the peer's own record, but makes sense to avoid any confusion.

PR #28, opened 2020-03-16
Index sessions by id and address

ajsutton:
## PR Description
Discv5 requires that when a node changes it's source address (ie roaming between networks) a new handshake should be initiated by sending a WHOAREYOU response.  To achieve this, `NodeSession` should be indexed by (node ID, IP, port) not just by node ID.

PONG messages are now able to use the remote socket address of the peer, rather than the address they are reporting in the ENR as required by the spec.

Also adds simple accessor methods to `NodeRecord` to get the TCP and UDP addresses via the `IdentitySchemaInterpreter` rather than duplicating that logic in multiple places.


src/test/java/org/ethereum/beacon/discovery/integration/DiscoveryIntegrationTest.java
```diff
@@ -84,13 +84,39 @@ public void shouldNotSuccessfullyPingBootnodeWhenNodeRecordIsNotSigned() throws
...
+    final DiscoverySystem node2 = createDiscoveryClient(keyPair, node1.getLocalNodeRecord());
+
+    // Communicate on the first port
+    waitFor(node2.ping(node1.getLocalNodeRecord()));
+    node2.stop();
+
+    // Then with the same node ID but a different port
+    final DiscoverySystem node2OnDifferentPort =
+        createDiscoveryClient(keyPair, node1.getLocalNodeRecord());
+    waitFor(node2OnDifferentPort.ping(node1.getLocalNodeRecord()));
```
cemozerr: Is .ping() supposed to throw an error? Otherwise, I can't see this test checking correctness.

ajsutton: If we don't get the PONG back correctly it will time out.  It's not really the ideal test - I'd like to be able to assert that it went through another handshake but can't at this level.  But its still useful as a basic sanity check that things do keep working (I managed to break that at one point in the development).

PR #33, opened 2020-04-02
Reuse a single thread for scheduling expirations

ajsutton:
## PR Description
Fixes a thread leak where a new scheduled executor was created for every `NodeSession` by reusing a single `ScheduledExecutor` across all `ExpirationScheduler` instances.

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/DiscoverySystemBuilder.java
+++ b/src/main/java/org/ethereum/beacon/discovery/DiscoverySystemBuilder.java
@@ -104,7 +104,7 @@ public class DiscoverySystemBuilder {
             localNodeRecordStore,
             privateKey,
             nodeRecordFactory,
-            schedulers.newSingleThreadDaemon("client-" + clientNumber),
+            schedulers.newSingleThreadDaemon("discovery-client-" + clientNumber),
             expirationSchedulerFactory);
 
     final DiscoveryTaskManager discoveryTaskManager =
@@ -113,7 +113,7 @@ public class DiscoverySystemBuilder {
             nodeTable,
             nodeBucketStorage,
             localNodeRecord,
-            schedulers.newSingleThreadDaemon("tasks-" + clientNumber),
+            schedulers.newSingleThreadDaemon("discovery-tasks-" + clientNumber),
             true,
             true,
             expirationSchedulerFactory);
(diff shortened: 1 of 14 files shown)


src/main/java/org/ethereum/beacon/discovery/DiscoverySystemBuilder.java
```diff
@@ -90,14 +93,19 @@ public DiscoverySystem build() {
...
+                new ThreadFactoryBuilder().setNameFormat("discovery-expiration-%d").build()));
     final DiscoveryManager discoveryManager =
         new DiscoveryManagerImpl(
             nodeTable,
             nodeBucketStorage,
             localNodeRecordStore,
             privateKey,
             nodeRecordFactory,
-            schedulers.newSingleThreadDaemon("client-" + clientNumber));
+            schedulers.newSingleThreadDaemon("client-" + clientNumber),
```
cemozerr, line 107: nit: Would it be better to have a Thread name that gives more context into what the thread is doing? 
```suggestion
            schedulers.newSingleThreadDaemon("client-expiration-scheduler-" + clientNumber),
```

ajsutton: Good idea, but that's not the expiration schedule thread, that line is unchanged except for the `,` at the end so we could pass the `expirationSchedulerFactory` through.  It's thread is `discovery-expiration-%d`.  I've prepended them with `discovery-` to make it clearer.

src/main/java/org/ethereum/beacon/discovery/schema/NodeSession.java
```diff
@@ -74,7 +74,8 @@ public NodeSession(
       NodeBucketStorage nodeBucketStorage,
       AuthTagRepository authTagRepo,
       Consumer<NetworkParcel> outgoingPipeline,
-      Random rnd) {
+      Random rnd,
+      ExpirationSchedulerFactory expirationSchedulerFactory) {
```
cemozerr, line 78: nit: Would it make more sense to do a dependency injection instead here? 
```suggestion
      ExpirationScheduler expirationScheduler) {
```

ajsutton: Yeah it probably does in that case - good spot.  Generally I've tried to be fairly low touch since the design was built with expiration schedulers being specific to a particular case, so just aimed to ensure that a single `ScheduledExecutorService` was shared between them all.

ajsutton:
With a single threaded scheduled executor, only one task can fire at the same time.  So if you schedule something in 60 seconds time and something else in 30 seconds time, the thread sits idle for 30 seconds, then executes the second task, then sits idle until 60 seconds is reached (30 seconds - execution time of the first task) and executes the first.

If you schedule two tasks in 60 seconds time, then yes the execute sequentially.  The tasks we're scheduling are a) very quick - typically just removing from a map and b) not time critical as they're just expiring things.

PR #36, opened 2020-04-14
Add custom field to Node Record

cemozerr:
## PR Description
Adds a custom field to Node Record object and API to update it through Discovery System.

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/schema/IdentitySchemaV4Interpreter.java
+++ b/src/main/java/org/ethereum/beacon/discovery/schema/IdentitySchemaV4Interpreter.java
@@ -4,6 +4,8 @@
 
 package org.ethereum.beacon.discovery.schema;
 
+import static org.ethereum.beacon.discovery.schema.NodeRecordBuilder.addCustomField;
+
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
 import java.net.InetAddress;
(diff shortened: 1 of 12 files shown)


src/main/java/org/ethereum/beacon/discovery/schema/IdentitySchemaV4Interpreter.java
```diff
@@ -116,4 +128,15 @@ public NodeRecord createWithNewAddress(
   private static InetAddress getInetAddress(final Bytes address) throws UnknownHostException {
     return InetAddress.getByAddress(address.toArrayUnsafe());
   }
+
+  private static Stream<EnrField> getAllFields(final NodeRecord nodeRecord) {
```
ajsutton, line 132: nit: `streamAllFields` tends to be the convention for things that return a `Stream`.

src/main/java/org/ethereum/beacon/discovery/schema/NodeRecordBuilder.java
```diff
@@ -70,6 +75,11 @@ static void addFieldsForUdpAddress(
     fields.add(new EnrField(isIpV6 ? EnrField.UDP_V6 : EnrField.UDP, udpPort));
   }
 
+  static void addCustomField(
+      final List<EnrField> fields, final String fieldName, final Bytes value) {
+    fields.add(new EnrField(fieldName, value));
+  }
```
ajsutton, line 81: Does this really need to be it's own method?

cemozerr: Not really but I wanted to make sure I adhere to the convention.

src/main/java/org/ethereum/beacon/discovery/schema/IdentitySchemaV4Interpreter.java
cemozerr: done.

PR #47, opened 2020-05-21
Handle Optionals safely

ajsutton:
## PR Description
Fixes the exception reported in https://github.com/PegaSysEng/teku/issues/1792

Handles node table lookups that don't find the entry in question (it may have been removed in the meantime safely) and ensure that the TaskManager executes recursive searches on it's task thread preserving the correct threading model.

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/storage/NodeTableImpl.java
+++ b/src/main/java/org/ethereum/beacon/discovery/storage/NodeTableImpl.java
@@ -166,9 +166,7 @@ public class NodeTableImpl implements NodeTable {
         upNodesOptional.ifPresent(
             upNodes -> {
               for (Bytes currentNodeId : upNodes.getEntries()) {
-                if (logLimit == 0) {
-                  getNode(currentNodeId).ifPresent(res::add);
-                } else if (Functions.logDistance(currentNodeId, nodeId) >= logLimit) {
+                if (logLimit != 0 && Functions.logDistance(currentNodeId, nodeId) >= logLimit) {
                   limitReached = true;
                   break;
                 } else {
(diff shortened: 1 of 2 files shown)


src/main/java/org/ethereum/beacon/discovery/storage/NodeTableImpl.java
```diff
@@ -163,35 +163,35 @@ private void loadNextBatch() {
...
+            upNodes -> {
+              for (Bytes currentNodeId : upNodes.getEntries()) {
+                if (logLimit == 0) {
+                  getNode(currentNodeId).ifPresent(res::add);
+                } else if (Functions.logDistance(currentNodeId, nodeId) >= logLimit) {
+                  limitReached = true;
+                  break;
+                } else {
+                  getNode(currentNodeId).ifPresent(res::add);
+                }
```
mbaxter, line 176: (nit) Can this be simplified?:
```suggestion
                if (logLimit != 0 && Functions.logDistance(currentNodeId, nodeId) >= logLimit) {
                  limitReached = true;
                  break;
                } else {
                  getNode(currentNodeId).ifPresent(res::add);
                }
```

src/main/java/org/ethereum/beacon/discovery/task/DiscoveryTaskManager.java
```diff
@@ -229,9 +230,11 @@ private void liveCheckTask() {
...
 
   public CompletableFuture<Void> searchForNewPeers() {
-    return new RecursiveLookupTask(
-            nodeTable, this::findNodes, RECURSIVE_SEARCH_QUERY_LIMIT, Bytes32.random())
-        .execute();
+    final RecursiveLookupTask task =
+        new RecursiveLookupTask(
+            nodeTable, this::findNodes, RECURSIVE_SEARCH_QUERY_LIMIT, Bytes32.random());
+    // We wind up with a CompletableFuture<CompletableFuture> so unwrap one level.
+    return scheduler.execute(task::execute).thenCompose(Function.identity());
```
mbaxter, line 237: Why move this into the scheduler?  Looks like `searchForPeers` is already running inside of the scheduler most of the time: https://github.com/PegaSysEng/discovery/blob/0e8b14fe55d983dfe845dddad91b4ce940b20b0a/src/main/java/org/ethereum/beacon/discovery/task/DiscoveryTaskManager.java#L155-L159

Maybe we need an internal method that gets scheduled at a fixed rate, and a separate public method that runs this in the scheduler?

ajsutton: True, good spot.

src/main/java/org/ethereum/beacon/discovery/storage/NodeTableImpl.java
ajsutton: Yep that works.

PR #67, opened 2020-08-28
hash cache(homeNodeId)

### LINKED ISSUE #66: hash(homeNodeId) can be cached
Each time a `RandomPacket` is received, the sourceNodeId must be calculated from the tag field.  This requires the hash(homeNodeId) to be calculated:

Call to `getSourceNodeId(homeNodeId)`
https://github.com/PegaSysEng/discovery/blob/840e90bea80699790eb750839be7402fe8b598b0/src/main/java/org/ethereum/beacon/discovery/pipeline/handler/UnknownPacketTagToSender.java#L49

hash(homeNodeId) seems to happen each time:
https://github.com/PegaSysEng/discovery/blob/840e90bea80699790eb750839be7402fe8b598b0/src/main/java/org/ethereum/beacon/discovery/packet/UnknownPacket.java#L58

This can be avoided if the hash(homeNodeId) is cached.

jrhea:
## PR Description
caching the hash(homeNodeId) so it doesn't have to be recalculated each time a tag is decoded on a RandomPacket
## Fixed Issue(s)
resolves #66

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/packet/UnknownPacket.java
+++ b/src/main/java/org/ethereum/beacon/discovery/packet/UnknownPacket.java
@@ -7,7 +7,6 @@ package org.ethereum.beacon.discovery.packet;
 import com.google.common.base.Preconditions;
 import java.util.Optional;
 import org.apache.tuweni.bytes.Bytes;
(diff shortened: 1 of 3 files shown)


ajsutton (approved):
LGTM.

PR #69, opened 2020-09-02
Add local ENR change listener

Nashatyrev:
## PR Description

Address the case when the client code needs to be notified on the local ENR updates. 

## Fixed Issue(s)

the part of https://github.com/PegaSysEng/teku/pull/2715

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
+++ b/src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
@@ -9,12 +9,20 @@ import org.apache.tuweni.bytes.Bytes;
 import org.ethereum.beacon.discovery.schema.NodeRecord;
 
 public class LocalNodeRecordStore {
+
   private volatile NodeRecord latestRecord;
   private final Bytes privateKey;
+  private final NodeRecordListener recordListener;
 
   public LocalNodeRecordStore(final NodeRecord record, final Bytes privateKey) {
-    latestRecord = record;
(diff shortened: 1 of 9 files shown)


src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
```diff
@@ -9,26 +9,37 @@
 import org.ethereum.beacon.discovery.schema.NodeRecord;
 
 public class LocalNodeRecordStore {
+
   private volatile NodeRecord latestRecord;
   private final Bytes privateKey;
+  private final NodeRecordListener recordListener;
 
   public LocalNodeRecordStore(final NodeRecord record, final Bytes privateKey) {
```
mbaxter, line 17: (nit) Might be a good idea to just have 1 constructor and require the listener argument.

src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
```diff
@@ -9,26 +9,37 @@
...
+    this.latestRecord = record;
     this.privateKey = privateKey;
+    this.recordListener = recordListener;
   }
 
   public NodeRecord getLocalNodeRecord() {
     return latestRecord;
   }
 
   public void onSocketAddressChanged(final InetSocketAddress newAddress) {
```
mbaxter, line 32: Probably worth some quick unit tests to verify the listener is invoked as expected.

PR #70, opened 2020-09-08
Add packet AuthTag RLP and minimal RandomPacket size validation

Nashatyrev:
## PR Description

- verify `authTag` RLP contains expected data: a single 12 bytes string 
- verify that `RandomPacket` meets the minimum size requirement 
- refactor `RandomPacket` and `MessagePacket` as they have the same `tag+authTag` header: extracted common `TaggedMessage` class
- add RandomPacketTest
- Fix the gradle IDEA bug (see for details here: https://github.com/PegaSysEng/teku/pull/1091)

## Fixed Issue(s)
Fix https://github.com/PegaSysEng/teku/issues/2648
Relates https://github.com/web3j/web3j/issues/1252

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/packet/MessagePacket.java
+++ b/src/main/java/org/ethereum/beacon/discovery/packet/MessagePacket.java
@@ -9,9 +9,6 @@ import org.apache.tuweni.bytes.MutableBytes;
 import org.ethereum.beacon.discovery.message.DiscoveryMessage;
 import org.ethereum.beacon.discovery.message.DiscoveryV5Message;
 import org.ethereum.beacon.discovery.util.Functions;
-import org.web3j.rlp.RlpDecoder;
-import org.web3j.rlp.RlpEncoder;
-import org.web3j.rlp.RlpString;
 
 /**
  * Used when handshake is completed as a {@link DiscoveryMessage} authenticated container
@@ -22,16 +19,9 @@ import org.web3j.rlp.RlpString;
  * message-pt = message-type || message-data</code>
  */
 public class MessagePacket extends AbstractPacket {
-  private MessagePacketDecoded decoded = null;
-
-  public MessagePacket(Bytes bytes) {
-    super(bytes);
-  }
 
   public static MessagePacket create(Bytes tag, Bytes authTag, Bytes messageCipherText) {
-    byte[] authTagBytesRlp = RlpEncoder.encode(RlpString.create(authTag.toArray()));
-    Bytes authTagEncoded = Bytes.wrap(authTagBytesRlp);
-    return new MessagePacket(Bytes.concatenate(tag, authTagEncoded, messageCipherText));
+    return new MessagePacket(new TaggedMessage(tag, authTag, messageCipherText));
   }
 
   public static MessagePacket create(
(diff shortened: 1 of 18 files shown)


src/main/java/org/ethereum/beacon/discovery/packet/TaggedMessage.java
```diff
@@ -0,0 +1,91 @@
...
+
+  private final Bytes tag;
+  private final Bytes authTag;
+  private final Bytes payload;
+
+  public static TaggedMessage decode(Bytes bytes) {
+    Bytes tag = Bytes.wrap(bytes.slice(0, TAG_SIZE));
+    Bytes authTagRlpBytes = bytes.slice(TAG_SIZE, RLP_AUTH_TAG_SIZE);
+    Bytes authTag = decodeSingleRlpString(authTagRlpBytes);
+    if (authTag.size() != 12) {
```
jrhea, line 34: Since you went to the trouble of defining constants, why not use `RLP_AUTH_TAG_SIZE -1`?  Or do something like this:

```java
  public static final int TAG_SIZE = 32;
  public static final int RLP_AUTH_TAG_PREFIX_SIZE = 1;
  public static final int AUTH_TAG_SIZE = 12;
  public static final int RLP_AUTH_TAG_SIZE = AUTH_TAG_SIZE + RLP_AUTH_TAG_PREFIX_SIZE;
  public static final int HEADER_SIZE = TAG_SIZE + RLP_AUTH_TAG_SIZE;

  private final Bytes tag;
  private final Bytes authTag;
  private final Bytes payload;

  public static TaggedMessage decode(Bytes bytes) {
    Bytes tag = Bytes.wrap(bytes.slice(0, TAG_SIZE));
    Bytes authTagRlpBytes = bytes.slice(TAG_SIZE, RLP_AUTH_TAG_SIZE);
    Bytes authTag = decodeSingleRlpString(authTagRlpBytes);
    if (authTag.size() != AUTH_TAG_SIZE) {
```

jrhea (commented):
This is definitely an improvement. This fix addresses the issue where the auth_tag isn't properly prefixed with `8c` by simply decoding the auth_tag bytes and making sure that 12 bytes are returned.  That's great. It works. Just one small nit:

- Perhaps consider using a predefined constant to check the size of the decoded RLP auth tag in the `decode()` method of `TaggedMessage`.  I left a more detailed comment inline

The other issue this PR addresses is how the Discovery service will reply to undersized `RANDOM PACKETS`.  It turns out that this requires a Discv5 spec change.  See @fjl comments in [lighthouse](https://github.com/sigp/lighthouse/issues/1568#issuecomment-683414834) and [prysm](https://github.com/prysmaticlabs/prysm/issues/7110#issuecomment-686942272) repos.

src/main/java/org/ethereum/beacon/discovery/packet/TaggedMessage.java
Nashatyrev: Oh, right, thanks 👍

Nashatyrev:
> The other issue this PR addresses is how the Discovery service will reply to undersized RANDOM PACKETS. It turns out that this requires a Discv5 spec change. See @fjl comments in lighthouse and prysm repos.

@jrhea  Right with the current spec there is no much sense to set the lower limit of the `RandomPacket` (you may always send a legit `Ping` instead which is smaller than `RandomPacket`)
I'd suggest to merge this PR and address the issue separately as soon as the spec changes are ready

jrhea:
> I'd suggest to merge this PR and address the issue separately as soon as the spec changes are ready

I agree.  LGTM.  I would hold off on merging until @ajsutton can review

jrhea (approved):

src/main/java/org/ethereum/beacon/discovery/packet/MessagePacket.java
```diff
@@ -45,66 +35,73 @@ public static MessagePacket create(
...
-              .getBytes());
-    }
-    return decoded.authTag;
+    return getTaggedMessage().getAuthTag();
+  }
+
+  public Bytes getHomeNodeId(Bytes destNodeId) {
+    return Functions.hash(destNodeId)
+        .xor(
+            decodedTaggedMessage.getTag(),
```
ajsutton, line 57: Should this go via `getTaggedMessage`?

src/main/java/org/ethereum/beacon/discovery/packet/TaggedMessage.java
```diff
@@ -0,0 +1,93 @@
...
+  public static final int AUTH_TAG_SIZE = 12;
+  public static final int RLP_AUTH_TAG_SIZE = AUTH_TAG_SIZE + RLP_AUTH_TAG_PREFIX_SIZE;
+  public static final int HEADER_SIZE = TAG_SIZE + RLP_AUTH_TAG_SIZE;
+
+  private final Bytes tag;
+  private final Bytes authTag;
+  private final Bytes payload;
+
+  public static TaggedMessage decode(Bytes bytes) {
+    Bytes tag = Bytes.wrap(bytes.slice(0, TAG_SIZE));
```
ajsutton, line 33: Should we assert the data is long enough before slicing it? I seem to recall we've had unhandled exceptions from this kind of thing before but it's possible decoding failures are handled elsewhere.

ajsutton: Ok, we do handle any exception further up - personally I'd probably still do a `checkState` to confirm the length for good measure but it will work either way.

ajsutton (approved):
LGTM.

src/main/java/org/ethereum/beacon/discovery/packet/MessagePacket.java
Nashatyrev: Good catch! Thanks!

PR #71, opened 2020-09-10
Fix discovery out of service

Nashatyrev:
## PR Description

- Address this attack https://github.com/PegaSysEng/teku/issues/2778 (small change: 9787550 )
- More refactorings with safer RLP decoding (larger change: 6cbf835)

The latter refactor adds a number of RLP checks, however a lot of restrictions left which are still not verified explicitly. For the comprehensive RLP verification we would need some scheme based decoding 

## Fixed Issue(s)
Fixes https://github.com/PegaSysEng/teku/issues/2778

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/packet/UnknownPacket.java
+++ b/src/main/java/org/ethereum/beacon/discovery/packet/UnknownPacket.java
@@ -29,10 +29,6 @@ public class UnknownPacket extends AbstractPacket {
     return new RandomPacket(getBytes());
   }
 
-  public TaggedMessage getTaggedMessage() {
-    return TaggedMessage.decode(getBytes());
-  }
-
   public WhoAreYouPacket getWhoAreYouPacket() {
     return new WhoAreYouPacket(getBytes());
   }
--- a/src/main/java/org/ethereum/beacon/discovery/packet/UnknownPacket.java
+++ b/src/main/java/org/ethereum/beacon/discovery/packet/UnknownPacket.java
@@ -29,6 +29,10 @@ public class UnknownPacket extends AbstractPacket {
     return new RandomPacket(getBytes());
   }
 
+  public TaggedMessage getTaggedMessage() {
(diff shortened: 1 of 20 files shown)


src/main/java/org/ethereum/beacon/discovery/packet/UnknownPacket.java
```diff
@@ -29,6 +29,10 @@ public RandomPacket getRandomPacket() {
     return new RandomPacket(getBytes());
   }
 
+  public TaggedMessage getTaggedMessage() {
```
mbaxter, line 32: doesn't look like this is used

src/main/java/org/ethereum/beacon/discovery/util/RlpUtil.java
```diff
@@ -8,19 +8,25 @@
...
 import org.web3j.rlp.RlpString;
+import org.web3j.rlp.RlpType;
 
 /**
  * Handy utilities used for RLP encoding and decoding and not fulfilled by {@link
  * org.web3j.rlp.RlpEncoder} and {@link RlpDecoder}
  */
 public class RlpUtil {
+  public static final int ANY_LEN = -1;
+  public static final int UINT64_LEN = -2;
```
mbaxter, line 28: doesn't look like this value is used anywhere

src/main/java/org/ethereum/beacon/discovery/util/RlpUtil.java
```diff
@@ -63,6 +69,118 @@ public static DecodedList decodeFirstList(Bytes data) {
     return new DecodedList(RlpDecoder.decode(data.slice(0, len).toArray()), data.slice(len));
   }
 
+  public static UInt64 itemToUInt(RlpType item) throws RlpDecodeException {
```
mbaxter, line 72: Looks like this is unused

mbaxter (approved):
LGTM

PR #80, opened 2020-11-18
When receiving invalid Handshake packet the session should be dropped

Nashatyrev:
## PR Description

When receiving invalid Handshake packet the session should be dropped I.e. another negotiation round (Random, WhoAreYou) needs to be performed.
Else the peer could be DoSed with a single invalid Handshake packet forcing to perform expensive `BLS` signature validation for free from attacker side 

## Fixed Issue(s)

When merged https://github.com/ConsenSys/teku/issues/2846

### DIFF
--- a/src/test/java/org/ethereum/beacon/discovery/DiscoveryManagerTest.java
+++ b/src/test/java/org/ethereum/beacon/discovery/DiscoveryManagerTest.java
@@ -3,8 +3,8 @@
  */
 package org.ethereum.beacon.discovery;
 
+import static java.time.Duration.ofSeconds;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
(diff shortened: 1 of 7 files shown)


src/test/java/org/ethereum/beacon/discovery/DiscoveryManagerTest.java
```diff
@@ -45,6 +57,64 @@ public void testRegularHandshake() throws Exception {
...
+            handshakeRawPacket.getMaskingIV(), malformedHandshakePacket, maskingKey);
+    TestMessage malformedMessage = attackerNode.createOutbound(malformedRawPacket, victimNode);
+
+    victimNode.deliver(malformedMessage); // Malformed Handshake
+    victimNode.deliver(malformedMessage); // Malformed Handshake
+    victimNode.deliver(validHandshakeMessage);
+
+    // victim node should drop the session on the first malformed handshake message
+    // and ignore any subsequent handshake messages
+    assertThatThrownBy(victimNode::nextOutbound);
```
mbaxter, line 115: This is a little confusing to read without any assertions on the error thrown.  Should probably add some assertions here, or have `nextOutbound` just return an `Optional`.

mbaxter (approved):
LGTM

PR #84, opened 2020-12-29
Implement rate limiter

cemozerr:
## PR Description
Use Netty's `ChannelTrafficShapingHandler` to limit the number of messages and protect against DoS attacks. Include the attack code for further testing.

In order to find a good limit for the number of read bytes/second, I've tested and included the CPU usage over time for the numbers given below. The attacker CPU usage has consistently been very close to 80%. 

[netty_traffic_shaping_1500_bytes_per_sec.txt](https://github.com/ConsenSys/discovery/files/5755567/netty_traffic_shaping_1500_bytes_per_sec.txt)
[netty_traffic_shaping_15000_bytes_per_sec.txt](https://github.com/ConsenSys/discovery/files/5755568/netty_traffic_shaping_15000_bytes_per_sec.txt)
[netty_traffic_shaping_150kb_per_sec.txt](https://github.com/ConsenSys/discovery/files/5755574/netty_traffic_shaping_150kb_per_sec.txt)
[netty_traffic_shaping_250kb_per_sec.txt](https://github.com/ConsenSys/discovery/files/5766436/netty_traffic_shaping_250kb_per_sec.txt)
[netty_traffic_shaping_500kb_per_sec.txt](https://github.com/ConsenSys/discovery/files/5755573/netty_traffic_shaping_500kb_per_sec.txt)
[netty_traffic_shaping_1mb_bytes_per_sec.txt](https://github.com/ConsenSys/discovery/files/5755617/netty_traffic_shaping_1mb_bytes_per_sec.txt)

Number of peers discovered in 5 minutes using 2 PegaSys mainnet bootnodes:
1mb: 758 
500kb: 1011
250kb: 865
150kb: 795
15000 bytes: 813


src/main/java/org/ethereum/beacon/discovery/DiscoveryManagerImpl.java
```diff
@@ -93,6 +96,7 @@ public DiscoveryManagerImpl(
             outgoingPipeline,
             expirationSchedulerFactory);
     incomingPipeline
+        .addHandler(rateLimiter)
```
Nashatyrev, line 99: I would consider moving the rate limiter even more close to the 'origin', e.g. here to the very head of Netty pipeline  https://github.com/ConsenSys/discovery/blob/2eedf4e98b6ffe12f455d74dca2a26eb50bf2c5b/src/main/java/org/ethereum/beacon/discovery/network/NettyDiscoveryServerImpl.java#L60

This way no our code would be affected by a possible naive DoS attack 

In the latter I would also check for existing Netty rate limit handlers (I thought I saw something like this)

src/main/java/org/ethereum/beacon/discovery/network/NettyDiscoveryServerImpl.java
```diff
@@ -57,6 +60,7 @@ public NettyDiscoveryServerImpl(InetSocketAddress listenAddress) {
               @Override
               public void initChannel(NioDatagramChannel ch) {
                 ch.pipeline()
+                    .addFirst(new ChannelTrafficShapingHandler(0, trafficReadLimit))
                     .addFirst(new LoggingHandler(LogLevel.TRACE))
```
ajsutton, line 64: We're now calling `addFirst` twice so I believe this setup means the `LoggingHandler` winds up added prior to the traffic shaping handler which probably isn't what we want.

src/main/java/org/ethereum/beacon/discovery/network/NettyDiscoveryServerImpl.java
```diff
@@ -57,6 +60,7 @@ public NettyDiscoveryServerImpl(InetSocketAddress listenAddress) {
               @Override
               public void initChannel(NioDatagramChannel ch) {
                 ch.pipeline()
+                    .addFirst(new ChannelTrafficShapingHandler(0, trafficReadLimit))
```
ajsutton, line 63: I'd suggest having a way to disable this entirely - e.g. by setting `trafficReadLimit` to 0.  I guess it could be set to Long.MAX_VALUE but would be nice to be able to skip adding the handler altogether.

src/test/java/org/ethereum/beacon/discovery/Attack.java
```diff
@@ -0,0 +1,69 @@
...
+import org.ethereum.beacon.discovery.packet.impl.RawPacketImpl;
+import org.ethereum.beacon.discovery.type.Bytes12;
+import org.ethereum.beacon.discovery.type.Bytes16;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+
+public class Attack {
+  @Test
+  @Disabled
+  void dos() throws Exception {
```
ajsutton, line 24: Not sure we should commit this - it's disabled so won't run and is one of those bits of code that's just likely to rot and wind up not working next time we actually go to run it anyway.

src/test/java/org/ethereum/beacon/discovery/app/DiscoveryTestServer.java
```diff
@@ -32,6 +32,8 @@
 
   public static void main(String[] args)
       throws InterruptedException, ExecutionException, TimeoutException {
+    long pid = ProcessHandle.current().pid();
+    System.out.println("process id: " + pid);
```
ajsutton, line 36: Probably should remove this now.

src/test/resources/log4j2.xml
```diff
@@ -26,13 +26,13 @@
...
   <Loggers>
-    <Logger name="org.ethereum.beacon.discovery" level="trace" additivity="false">
+    <Logger name="org.ethereum.beacon.discovery" level="error" additivity="false">
       <AppenderRef ref="Console"/>
     </Logger>
 
     <Logger name="io.netty.util.internal" level="INFO"/>
 
-    <Root level="trace">
+    <Root level="error">
```
ajsutton, line 35: Probably shouldn't commit these changes either.

Nashatyrev (commented):
Looks good to me in general! `250 kb/sec` looks like a pretty reasonable limit.

src/main/java/org/ethereum/beacon/discovery/DiscoverySystemBuilder.java
```diff
@@ -135,7 +141,8 @@ private void createDefaults() {
                         .orElseThrow(
                             () ->
                                 new IllegalArgumentException(
-                                    "Local node record must contain an IP and UDP port"))));
+                                    "Local node record must contain an IP and UDP port")),
```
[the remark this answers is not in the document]
cemozerr: Done.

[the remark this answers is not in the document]
cemozerr: Done.

src/main/java/org/ethereum/beacon/discovery/network/NettyDiscoveryServerImpl.java
cemozerr: Fixed.

src/main/java/org/ethereum/beacon/discovery/network/NettyDiscoveryServerImpl.java
cemozerr: Done.

src/test/java/org/ethereum/beacon/discovery/Attack.java
cemozerr: Makes sense. Deleted.

src/test/java/org/ethereum/beacon/discovery/app/DiscoveryTestServer.java
cemozerr: Done.

src/test/resources/log4j2.xml
cemozerr: Done.

ajsutton (approved):
LGTM.

PR #85, opened 2021-01-19
Provide callback to change node record on new address

cemozerr:
## PR Description
Provide callback to change node record on new address

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/DiscoverySystemBuilder.java
+++ b/src/main/java/org/ethereum/beacon/discovery/DiscoverySystemBuilder.java
@@ -164,7 +164,9 @@ public class DiscoverySystemBuilder {
     localNodeRecordStore =
         requireNonNullElseGet(
             localNodeRecordStore,
-            () -> new LocalNodeRecordStore(localNodeRecord, privateKey, localNodeRecordListener, newAddressListener));
+            () ->
+                new LocalNodeRecordStore(
+                    localNodeRecord, privateKey, localNodeRecordListener, newAddressListener));
     expirationSchedulerFactory =
         requireNonNullElseGet(
             expirationSchedulerFactory,
--- a/src/main/java/org/ethereum/beacon/discovery/DiscoverySystemBuilder.java
+++ b/src/main/java/org/ethereum/beacon/discovery/DiscoverySystemBuilder.java
@@ -110,8 +110,8 @@ public class DiscoverySystemBuilder {
     return this;
   }
 
-  public DiscoverySystemBuilder newAddressListener(final NewAddressHandler listener) {
-    this.newAddressHandler = listener;
+  public DiscoverySystemBuilder newAddressHandler(final NewAddressHandler handler) {
+    this.newAddressHandler = handler;
     return this;
   }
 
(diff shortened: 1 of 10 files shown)


src/main/java/org/ethereum/beacon/discovery/DiscoverySystemBuilder.java
```diff
@@ -47,7 +48,8 @@
   private final NodeRecordFactory nodeRecordFactory = NodeRecordFactory.DEFAULT;
   private Database database;
   private Schedulers schedulers;
-  private NodeRecordListener localNodeRecordListener = (a, b) -> {};
+  private NodeRecordListener localNodeRecordListener = NodeRecordListener.NOOP;
+  private NewAddressListener newAddressListener = NewAddressListener.NOOP;
```
ajsutton, line 52: nit: Probably better called `NewAddressHandler`.

src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
```diff
@@ -28,8 +30,8 @@ public NodeRecord getLocalNodeRecord() {
   public void onSocketAddressChanged(final InetSocketAddress newAddress) {
     NodeRecord oldRecord = this.latestRecord;
     NodeRecord newRecord = oldRecord.withNewAddress(newAddress, privateKey);
-    this.latestRecord = newRecord;
-    recordListener.recordUpdated(oldRecord, newRecord);
+    newAddressListener.newAddress(oldRecord, newRecord)
+            .ifPresent(record -> this.latestRecord = record);
```
ajsutton: We still need to notify the `recordListener` if the record is updated.  The two functions should stay independent - one decides if/what to change, the other is notified of any changes.  Otherwise the Teku side will have to duplicate code to ensure the disk storage is updated for both events instead of being able to do it in one place.

cemozerr: Good catch.

src/main/java/org/ethereum/beacon/discovery/DiscoverySystemBuilder.java
cemozerr: Done.

src/test/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStoreTest.java
```diff
@@ -34,8 +35,10 @@ public Update(NodeRecord oldRec, NodeRecord newRec) {
     List<Update> listenerCalls = new ArrayList<>();
     LocalNodeRecordStore recordStore =
         new LocalNodeRecordStore(
-            nodeRecord1, nodePair1.getPrivateKey(), (o, n) -> listenerCalls.add(new Update(o, n)));
-
+            nodeRecord1,
+            nodePair1.getPrivateKey(),
+            (o, n) -> listenerCalls.add(new Update(o, n)),
+            (o, n) -> Optional.of(n));
```
ajsutton, line 41: nit: Can just use the `NOOP` instance here I think.

ajsutton (approved):
LGTM.

src/test/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStoreTest.java
cemozerr: Done.

PR #90, opened 2021-02-23
Switch publishing to cloudsmith

ajsutton:
## PR Description
Switch publishing to cloudsmith instead of bintray.


mbaxter (approved):
LGTM

PR #98, opened 2021-03-23
Generate unique AES/GCM nonce per message

Nashatyrev:
## PR Description

Implementing missing part of the Discv5 spec:
> The number of messages which can be encrypted with a certain session key is limited because encryption of each message requires a unique nonce for AES-GCM. In addition to the keys, the session cache must also keep track of the count of outgoing messages to ensure the uniqueness of nonce values. Since the wire protocol uses 96 bit AES-GCM nonces, it is strongly recommended to generate them by encoding the current outgoing message count into the first 32 bits of the nonce and filling the remaining 64 bits with random data generated by a cryptographically secure random number generator.

- Generate nonce on every outbound packet following spec recommendations 
- Get rid of `NonceRepository` in favour of `NodeSessionManager`


src/main/java/org/ethereum/beacon/discovery/schema/NonceGenerator.java
```diff
@@ -0,0 +1,28 @@
...
+  private Bytes randomPart;
+
+  @Override
+  public synchronized Bytes12 apply(Random random) {
+    if (counter == counterEnd) {
+      randomPart = Bytes.random(RANDOM_PART_SIZE, random);
+      counter = random.nextInt();
+      counterEnd = counter;
+    }
+    counter++;
```
ajsutton, line 25: So the aim here is to generate a new randomPart each time counter wraps the whole way around an int value? That works in terms of tracking uniqueness which I think is all we need but just wanting to double check I'm understanding this right.

ajsutton (approved):
Assuming my understanding of the nonce generation is correct this LGTM.

src/main/java/org/ethereum/beacon/discovery/schema/NonceGenerator.java
Nashatyrev: Yep, this is exactly what the spec recommends: 
> Since the wire protocol uses 96 bit AES-GCM nonces, it is strongly recommended to generate them by encoding the current outgoing message count into the first 32 bits of the nonce and filling the remaining 64 bits with random data generated by a cryptographically secure random number generator.

PR #100, opened 2021-04-02
Added NodeRecordFactory.fromEnr method

### LINKED ISSUE #99: Add a new method, NodeRecordFactory.fromEnr(String enr)
### Description
Add a new method, `NodeRecordFactory.fromEnr(String enr)`, that will parse a string (represented as `enr:enrBase64`) and convert it to an `NodeRecord`. `NodeRecordFactory.fromEnr(String enr)` should be very similar to the existing method `NodeRecordFactory.fromBase64(String enrBase64)`, the only difference is that the new method should handle the `enr:` prefix.

The need for this new method revealed itself while fixing [teku issue 3583](https://github.com/ConsenSys/teku/issues/3583).

BogdanStirbat:
## PR Description
Added a new method, `NodeRecordFactory.fromEnr(String enr)`, that automatically handles the enr: prefix.

## Fixed Issue(s)

Fixes #99

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/schema/NodeRecordFactory.java
+++ b/src/main/java/org/ethereum/beacon/discovery/schema/NodeRecordFactory.java
@@ -63,7 +63,7 @@ public class NodeRecordFactory {
   }
 
   public NodeRecord fromEnr(String enr) {
-    return fromBase64(enr.substring("enr:".length()));
+    return fromBase64(enr.startsWith("enr:") ? enr.substring("enr:".length()) : enr);
   }
 
   public NodeRecord fromBytes(Bytes bytes) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/ethereum/beacon/discovery/schema/NodeRecordFactory.java
```diff
@@ -62,6 +62,10 @@ public NodeRecord fromBase64(String enrBase64) {
     return fromBytes(Base64.getUrlDecoder().decode(enrBase64));
   }
 
+  public NodeRecord fromEnr(String enr) {
+    return fromBase64(enr.substring("enr:".length()));
```
ajsutton, line 66: I'd probably keep the `enr:` prefix as optional (so do the `.startsWith("enr:')` check like Teku does. Otherwise we'll get a really cryptic error message about invalid base64 if the prefix is missing.

BogdanStirbat: I understand. I did the check like Teku does.

ajsutton (approved):
Thanks LGTM.   I'll tag a release so we can start taking advantage of this in Teku (there aren't a lot of changes to discovery so not sure when the next release would be otherwise).

ajsutton:
Discovery 0.4.6 is now out including this change.

PR #101, opened 2021-04-03
Remove the context from build job so it works for external contributors

ajsutton:
## PR Description
Remove the docker read only context from the `build` task as external contributors aren't authorised to access it causing builds not to run.  Docker image downloads may wind up slightly slower but that's not a big deal for discovery.


macfarla (approved):
LGTM

PR #102, opened 2021-04-19
Remove docker hub authentication

ajsutton:
## PR Description
So that builds work for external contributors, remove the docker hub credentials and context.  Downloads of docker images may wind up being throttled but given the short build times and infrequent changes that's unlikely to be an issue for this project.


rolfyone (approved):
LGTM once building

Issue #103, opened 2021-05-18
Investigate lack of peers reported by MainNet bootnodes


ajsutton:
### Description
The Teku MainNet bootnode reports far fewer peers than other bootnodes with most distances having no reported peers.

Our bootnode ENR: `enr:-KG4QJRlj4pHagfNIm-Fsx9EVjW4rviuZYzle3tyddm2KAWMJBDGAhxfM2g-pDaaiwE8q19uvLSH4jyvWjypLMr3TIcEhGV0aDKQ9aX9QgAAAAD__________4JpZIJ2NIJpcIQDE8KdiXNlY3AyNTZrMaEDhpehBDbZjM_L9ek699Y7vhUJ-eAdMyQW_Fil522Y0fODdGNwgiMog3VkcIIjKA`

For exploratory testing I've been using a new class in the discovery test source tree:
```
package org.ethereum.beacon.discovery;

import static org.ethereum.beacon.discovery.util.Functions.PRIVKEY_SIZE;

import java.util.ArrayList;
import java.util.List;
import java.util.Optional;
import java.util.Random;
import org.apache.tuweni.bytes.Bytes;
import org.ethereum.beacon.discovery.schema.NodeRecord;
import org.ethereum.beacon.discovery.schema.NodeRecordBuilder;
import org.ethereum.beacon.discovery.schema.NodeRecordFactory;
import org.ethereum.beacon.discovery.util.Functions;
import org.ethereum.beacon.discovery.util.Utils;
import org.web3j.crypto.ECKeyPair;

public class Playground {
  public static void main(String[] args) {
    ECKeyPair keyPair = Functions.generateECKeyPair(new Random(1));
    final Bytes privateKey =
        Bytes.wrap(Utils.extractBytesFromUnsignedBigInt(keyPair.getPrivateKey(), PRIVKEY_SIZE));

    final DiscoverySystem system =
        new DiscoverySystemBuilder()
            .listen("0.0.0.0", 9000)
            .privateKey(privateKey)
            .localNodeRecord(
                new NodeRecordBuilder()
                    .privateKey(privateKey)
                    .address("180.150.110.29", 9000)
                    .seq(0)
                    .build())
            .newAddressHandler(
                (oldRecord, proposedRecord) -> {
                  System.out.println("Propsing address: " + proposedRecord);
                  return Optional.of(proposedRecord);
                })
            .build();
    final NodeRecord efBootnode =
        NodeRecordFactory.DEFAULT.fromEnr(
[...]

PR #104, opened 2021-06-03
Increase the number of packets allowed per message

rolfyone:
A lot of messages were being discarded because of the packet restriction, which limits the nodes we have in our tables, so increasing this should help with finding nodes.

partially addresses #103

Signed-off-by: Paul Harris <paul.harris@consensys.net>


SangameswaranRS:
Should the MAX_NODES_PER_MESSAGE = 4 and total =16 be changed also in FindNodeHandler ?

This change only handles the client side. As we have restricted nodes in node bucket(k=16), this change might help  only if we are querying for multiple buckets in List of distances that can be provided with a FIND_NODES request.

Please correct me if I'm wrong.

rolfyone:
> Should the MAX_NODES_PER_MESSAGE = 4 and total =16 be changed also in FindNodeHandler ?
> 
> This change only handles the client side. As we have restricted nodes in node bucket(k=16), this change might help only if we are querying for multiple buckets in List of distances that can be provided with a FIND_NODES request.
> 
> Please correct me if I'm wrong.

I was querying against an eth2 bootnode and was running up against this limit when the node would attempt to send back up to 9 packets in a message, so I guess ultimately I was acting as a client in this instance...

It appears that the limit is an arbitrary one and not set by the spec, but I could definitely be wrong, I've read the spec over but am still coming up to speed, i'd definitely be interested in any insights that anyone may have...

The issue I was investigating is why teku bootnodes don't seem to have a lot of peers, and this code basically drops all responses with more than 5 packets, meaning we never import the responses into our tables, if that adds any context...

SangameswaranRS:
@rolfyone , thanks for the explanation,
Basically what NODES packet we receive is sent from the Handler in pipeline, as the handler itself was restricted to 4 Max packets, Im confused so as how would we receive > 4 packets. 

Are the constraints different for an ETH boot node?
Just curious. Please correct if Im wrong.
Perhaps the authors can give a better explanation. @ajsutton

rolfyone:
I'm still putting it all together... I'm finding it hard to get any kind of commitment on limits in the spec, all I know is i was reading from a bootnode and was getting a lot of messages containing 9 packets, and they get dumped because they're more than what was deemed valid. I'll keep digging.

ajsutton:
I'm not the original author of this code - I just happened to be the one that imported it into git. 

I believe the original 5 packet response is because previously you could only request one distance per request and given the maximum size of an ENR, the maximum of 16 nodes per k-bucket and the maximum size of packets, you should be able to fit any response into 4 packets and 5 gives a little wiggle room.  However the spec now allows requesting multiple distances at once which may require more packets to return.

We want some limit in there so that we don't have to keep processing responses forever but the limit of what we choose to receive doesn't have to be the same as what we choose to send.  It looks like the EF bootnodes (using the go implementation) don't pack ENRs into packets as well as they could so wind up sending more packets than we would for the same response.

ajsutton:
And yes, it may also make sense to review the limit we place on the number of records we send in responses bu that's a separate issue.  It's probably good to keep that fairly small since it's a case where a small request generates a larger response - the requester has to go through the handshake process so it shouldn't be an amplification attack risk, but we still don't want to let the imbalance be too great or it becomes a DOS vector.

ajsutton (approved):

PR #105, opened 2021-06-07
Simplify storage system

ajsutton:
## PR Description
Simplify the storage system to make the code easier to understand and make debugging easier.

Only in-memory storage is ever used so avoid the need for serializing and restoring data.
Remove a lot of indirection.


mbaxter (approved):
LGTM

PR #106, opened 2021-06-08
Remove home node from NodeTable

ajsutton:
## PR Description
`NodeTable` doesn't actually need a reference to the home node so remove it and simplify things.


rolfyone (approved):
LGTM

PR #107, opened 2021-06-08
Include our seq no in PONG messages, not remote peer's

ajsutton:
## PR Description
When responding to PING messages, the PONG response should include our local ENR sequence number, not the remote peers.

This enables remote peers to detect when they need to request an updated ENR from us.


rolfyone (approved):
LGTM once CI passes

PR #108, opened 2021-06-08
Get latest home enr

ajsutton:
## PR Description
When requesting distance 0 from the bucket storage, always return the latest local node info rather than a previously stored version.

### DIFF
--- a/src/test/java/org/ethereum/beacon/discovery/DiscoveryInteropTest.java
+++ b/src/test/java/org/ethereum/beacon/discovery/DiscoveryInteropTest.java
@@ -114,6 +114,6 @@ public class DiscoveryInteropTest {
     // 1 sent findnodes to 2, received only (2) in answer
     // 1 added 2 to its nodeBuckets, because its now checked, but not before
     assertThat(nodeBucketStorage1.getNodeRecords(distance1To2).map(NodeRecord::getNodeId))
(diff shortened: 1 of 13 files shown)


src/test/java/org/ethereum/beacon/discovery/DiscoveryInteropTest.java
```diff
@@ -105,14 +108,12 @@ public void testInterop() throws Exception {
...
     assertTrue(authPacketSent1to2.await(1, TimeUnit.SECONDS));
     Thread.sleep(1000);
     // 1 sent findnodes to 2, received only (2) in answer
     // 1 added 2 to its nodeBuckets, because its now checked, but not before
-    NodeBucket bucketAt1With2 = nodeBucketStorage1.get(distance1To2).get();
-    Assertions.assertEquals(1, bucketAt1With2.size());
-    Assertions.assertEquals(
-        nodeRecord2.getNodeId(), bucketAt1With2.getNodeRecords().get(0).getNode().getNodeId());
+    assertThat(nodeBucketStorage1.getNodeRecords(distance1To2).map(NodeRecord::getNodeId))
+        .containsExactlyInAnyOrder(nodeRecord2.getNodeId());
```
rolfyone, line 117: nit: can probably just use `containsExactly` given the list is 1 element...

rolfyone (approved):
LGTM

PR #109, opened 2021-06-08
Update node table when newer ENRs received from FINDNODES

ajsutton:
## PR Description
When a `NODES` message is received that contains a newer ENR for a previously discovered node, update it's record in the node table.

Note that it is not updated in the bucket storage because we haven't validated the liveness of the new ENR yet (that's done through separate liveness checks).


rolfyone (approved):
LGTM

src/main/java/org/ethereum/beacon/discovery/message/handler/NodesHandler.java
```diff
@@ -50,13 +52,29 @@ public void handle(NodesMessage message, NodeSession session) {
...
+      nodeTable.save(nodeRecordInfo);
+
+      if (session.getNodeId().equals(nodeId)) {
+        // Node sent us a new version of their own ENR, update the session.
+        session.updateNodeRecord(nodeRecordV5);
+      }
+    }
+  }
+
+  private boolean isUpdated(
```
SangameswaranRS, line 74: nit: isUpdateRequired

PR #111, opened 2021-06-22
Add new implementation of k-bucket storage

ajsutton:
## PR Description
Add a new implementation of k-bucket storage that also handles the scheduling of liveness checks for nodes in those buckets.

Currently this isn't actually plugged in anywhere and the `LivenessChecker` is just an interface with no implementation. However this shows the most critical logic around maintaining the node table entries.

## Fixed Issue(s)
Part of #103

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/storage/KBucket.java
+++ b/src/main/java/org/ethereum/beacon/discovery/storage/KBucket.java
@@ -1,0 +1,162 @@
+/*
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package org.ethereum.beacon.discovery.storage;
+
+import java.time.Clock;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import org.ethereum.beacon.discovery.schema.NodeRecord;
+import org.ethereum.beacon.discovery.storage.KBuckets.LivenessChecker;
+
+public class KBucket {
+
+  static final int K = 16;
+
+  private final LivenessChecker livenessChecker;
+  private final Clock clock;
+
+  private final List<BucketEntry> nodes = new ArrayList<>();
+
+  private Optional<BucketEntry> pendingNode = Optional.empty();
+
+  public KBucket(final LivenessChecker livenessChecker, final Clock clock) {
+    this.livenessChecker = livenessChecker;
+    this.clock = clock;
+  }
+
+  public List<NodeRecord> getAllNodes() {
+    return nodes.stream().map(BucketEntry::getNode).collect(Collectors.toList());
+  }
+
+  public List<NodeRecord> getLiveNodes() {
+    return nodes.stream()
+        .filter(node -> node.isLive())
+        .map(BucketEntry::getNode)
+        .collect(Collectors.toList());
+  }
+
+  public Optional<NodeRecord> getPendingNode() {
+    return pendingNode.map(BucketEntry::getNode);
+  }
+
+  public void offer(final NodeRecord node) {
+    performMaintenance();
+    getEntry(node)
+        .ifPresentOrElse(
+            existing -> updateExistingRecord(existing, node), () -> offerNewNode(node));
+  }
+
+  private void updateExistingRecord(final BucketEntry existing, final NodeRecord newRecord) {
+    if (existing.getNode().getSeq().compareTo(newRecord.getSeq()) >= 0) {
+      // New record isn't actually newer, do nothing.
+      return;
+    }
+    nodes.remove(existing);
+    final BucketEntry newEntry = new BucketEntry(livenessChecker, newRecord);
+    nodes.add(newEntry);
+    newEntry.checkLiveness(clock.millis());
+  }
+
+  private void offerNewNode(final NodeRecord node) {
+    if (isFull()) {
+      getLastNode().checkLiveness(clock.millis());
+      if (pendingNode.isEmpty()) {
+        livenessChecker.checkLiveness(node);
+      }
+    } else {
+      final BucketEntry newEntry = new BucketEntry(livenessChecker, node);
+      nodes.add(newEntry);
+      newEntry.checkLiveness(clock.millis());
+    }
+  }
+
+  public void onNodeContacted(final NodeRecord node) {
+    getEntry(node)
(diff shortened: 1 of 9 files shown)


src/main/java/org/ethereum/beacon/discovery/storage/KBucket.java
```diff
@@ -0,0 +1,162 @@
...
+    final BucketEntry newEntry = new BucketEntry(livenessChecker, newRecord);
+    nodes.add(newEntry);
+    newEntry.checkLiveness(clock.millis());
+  }
+
+  private void offerNewNode(final NodeRecord node) {
+    if (isFull()) {
+      getLastNode().checkLiveness(clock.millis());
+      if (pendingNode.isEmpty()) {
+        livenessChecker.checkLiveness(node);
```
rolfyone, line 67: would the intent of checking liveness here be to set the pendingNode if the incoming `node` is live? if that's the intent, it seems we might want to `pendingNode = new BucketEntry(livenessChecker, node);` here...

src/main/java/org/ethereum/beacon/discovery/storage/KBucket.java
```diff
@@ -0,0 +1,162 @@
...
+
+  public void onNodeContacted(final NodeRecord node) {
+    getEntry(node)
+        .ifPresentOrElse(
+            existingEntry -> {
+              // Move to the start of the bucket
+              nodes.remove(existingEntry);
+              nodes.add(0, existingEntry.withLastConfirmedTime(clock.millis()));
+              performMaintenance();
+            },
```
rolfyone, line 84: there's not a LRU cache we can make use of? or just more efficient to maintain our own list?

src/main/java/org/ethereum/beacon/discovery/storage/KBucket.java
```diff
@@ -0,0 +1,162 @@
...
+   *
+   * <p>a. remove it from the bucket.
+   *
+   * <p>b. if there is a pending node, insert it into the bucket (at appropriate position based on
+   * when it was last confirmed as live)
+   */
+  public void performMaintenance() {
+    if (nodes.isEmpty()) {
+      return;
+    }
```
rolfyone, line 122: we might want to allow somehow getting into a state where nodes is empty, and we have a pending...

in that case we don't want to return here, we would just want to move past the nodes maintenance and onto the pending logic...

It would be a little sad skipping if we have a pending node that's valid and we just never insert it because the current list size is 0...

src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java
```diff
@@ -0,0 +1,87 @@
...
+  private Optional<KBucket> getBucket(final int distance) {
+    return Optional.ofNullable(buckets.get(distance));
+  }
+
+  public synchronized void offer(NodeRecord node) {
+    final int distance = Functions.logDistance(homeNodeId, node.getNodeId());
+    if (distance > MAXIMUM_BUCKET) {
+      // Distance too great, ignore.
+      return;
+    }
```
rolfyone, line 57: if it's outside the bucket max should it be even in our list? I'm not sure I fully understand why we might store a node that is > 256 distance...

src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Called when we have confirmed the liveness of a node by sending it a request and receiving a
+   * valid response back. Must only be called for requests we initiate, not incoming requests from
+   * the peer.
+   *
+   * @param node the node for which liveness was confirmed.
+   */
+  public synchronized void onNodeContacted(NodeRecord node) {
+    final int distance = Functions.logDistance(homeNodeId, node.getNodeId());
+    getOrCreateBucket(distance).onNodeContacted(node);
```
rolfyone, line 71: if we're concerned about distance we might need to check we're inside max distance here too?

src/main/java/org/ethereum/beacon/discovery/storage/KBucket.java
ajsutton: The intent of the pending node is to hold one entry which is confirmed to be live and thus ready to be inserted immediately should a node be removed. So as we discover nodes that don't fit, we test their liveness but only actually store them as the pending node (and stop pinging new nodes) once we hear back from them (which comes via `onNodeContacted`).

Might rename `onNodeContacted` to `onLivenessConfirmed` actually.

src/main/java/org/ethereum/beacon/discovery/storage/KBucket.java
ajsutton: It's not strictly a LRU cache - we need to be quite deliberate about when the node is moved to the front of the queue. We might access and even update the entry for a number of reasons like recording when it was last pinged, which aren't a trigger for moving the node to the front of the bucket as it doesn't confirm the node is live.

src/main/java/org/ethereum/beacon/discovery/storage/KBucket.java
ajsutton: I'm pretty sure it's never possible to have an empty nodes list and a pending node (and it would definitely be a bug if it happened) but yes, seems reasonable to always check the pending nodes here just in case.

src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java
ajsutton: This is a bit of a quirk because of the way `NodeRecord` is implemented.  The distance is the log distance so is actually the length of the common prefix of the two node IDs.  Since NodeIDs are a hash they are always 32 bytes and thus the greatest possible distance is 256, but `NodeRecord` uses `Bytes` not `Bytes32` as the type and has a pluggable system for calculating the node ID so we don't have any guarantee that it won't be longer than 32 bytes. So we wind up needing this check.

src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java
ajsutton: Yes good spot.

rolfyone (approved):
LGTM

PR #115, opened 2021-06-30
Don't assume that nodes are in the node table when sending requests

ajsutton:
## PR Description
Use the ENR supplied when initiating the request rather than requiring the peer to be present in the node table.
Still looks up the record from the node table if it is unknown (eg incoming requests).

Part of #111


rolfyone (approved):
LGTM

PR #117, opened 2021-07-02
Return found nodes

ajsutton:
## PR Description
Return the nodes found by a `findNodes` or `searchForNewPeers` request so they can be utilised even if they didn't get added to the node table (eg because it was full).

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/pipeline/info/FindNodeResponseHandler.java
+++ b/src/main/java/org/ethereum/beacon/discovery/pipeline/info/FindNodeResponseHandler.java
@@ -3,30 +3,112 @@
  */
 package org.ethereum.beacon.discovery.pipeline.info;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.tuweni.bytes.Bytes;
 import org.ethereum.beacon.discovery.message.NodesMessage;
+import org.ethereum.beacon.discovery.schema.NodeRecord;
+import org.ethereum.beacon.discovery.schema.NodeRecordInfo;
+import org.ethereum.beacon.discovery.schema.NodeSession;
+import org.ethereum.beacon.discovery.storage.NodeTable;
+import org.ethereum.beacon.discovery.util.Functions;
 
 public class FindNodeResponseHandler implements MultiPacketResponseHandler<NodesMessage> {
+  private static final Logger logger = LogManager.getLogger(FindNodeResponseHandler.class);
   private static final int MAX_TOTAL_PACKETS = 16;
+  private final List<NodeRecord> foundNodes = new ArrayList<>();
+  private final Collection<Integer> distances;
   private int totalPackets = -1;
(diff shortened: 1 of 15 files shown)


src/main/java/org/ethereum/beacon/discovery/pipeline/info/FindNodeResponseHandler.java
```diff
@@ -3,30 +3,112 @@
...
+    return existingRecord.isEmpty()
+        || existingRecord.get().getNode().getSeq().compareTo(newRecord.getSeq()) < 0;
+  }
+
+  private boolean isValid(final NodeRecord record) {
+    if (!record.isValid()) {
+      logger.debug("Rejecting invalid node record {}", record);
+      return false;
+    }
+    return true;
```
rolfyone, line 99: nit: I'd return `record.isValid()`, and use the condition just as a decision to debug...

rolfyone: although that admittedly does introduce a chance to have a change in value, maybe that's why this was done this way..

src/test/java/org/ethereum/beacon/discovery/pipeline/info/FindNodeResponseHandlerTest.java
```diff
@@ -0,0 +1,125 @@
...
+    assertThat(handler.handleResponseMessage(message, session)).isTrue();
+
+    final NodeRecordInfo nodeRecordInfo = NodeRecordInfo.createDefault(updatedRecord);
+    verify(nodeTable).save(nodeRecordInfo);
+    verify(session, never()).updateNodeRecord(any());
+    verify(session, never()).putRecordInBucket(any());
+  }
+
+  @Test
+  public void shouldUpdateNodeRecordInSessionWhenSelfEnrReturnedWithHigherSeqNum() {
```
rolfyone, line 79: `whenOwnEnrReturned`...

rolfyone (approved):
LGTM

src/main/java/org/ethereum/beacon/discovery/pipeline/info/FindNodeResponseHandler.java
ajsutton: Calling `isValid` requires performing potentially expensive cryptographic operations, hence the single call.  Potentially validity could be cached but it currently isn't.

src/test/java/org/ethereum/beacon/discovery/pipeline/info/FindNodeResponseHandlerTest.java
ajsutton: I meant to just rename this but forgot to actually push the change.  Self ENR sounds weird at times but it is actually the term used for the node's local ENR in most places.

src/main/java/org/ethereum/beacon/discovery/pipeline/info/FindNodeResponseHandler.java
rolfyone: true, but we can still call once by 
```
boolean result = record.isValid();
if (! result) ...

return isValid;
```

PR #118, opened 2021-07-21
Remove node table

ajsutton:


src/main/java/org/ethereum/beacon/discovery/schema/NodeSession.java
```diff
@@ -334,24 +310,30 @@ private synchronized RequestInfo clearRequestInfo(Bytes requestId) {
...
 
-  public synchronized Bytes getIdNonce() {
-    return idNonce;
+  public NodeRecord getHomeNodeRecord() {
+    return localNodeRecordStore.getLocalNodeRecord();
   }
 
-  public synchronized void setIdNonce(Bytes idNonce) {
-    this.idNonce = idNonce;
+  // TODO: Add tests that this updates internal record only when matching and seq higher
```
rolfyone, line 321: intentionally not actioning now? just a sanity check

src/test/java/org/ethereum/beacon/discovery/TestUtil.java
```diff
@@ -125,4 +129,46 @@ public static boolean isFieldsEqual(
...
+  public static void waitFor(final ThrowingRunnable assertion) throws Exception {
+    int attempts = 0;
+    while (true) {
+      try {
+        assertion.run();
+        return;
+      } catch (Throwable t) {
+        if (attempts < 60) {
+          attempts++;
+          Thread.sleep(1000);
```
rolfyone, line 151: for future it'd be nice to have a stub timer and not have to thread.sleep like this maybe, but its probably not worth doing now... just a thought.

rolfyone (approved):
LGTM

src/test/java/org/ethereum/beacon/discovery/TestUtil.java
ajsutton: This is an actual waiter for tests that are actually using multiple threads so it really does need to actually wait.  I think it's actually only used by `DiscoveryIntegrationTest` in the end and used to be a private method there but got moved because it looked like it would be useful elsewhere. Didn't move it back because it is a handy method to have around.

PR #119, opened 2021-07-22
During recursive lookups, use nodes found by previous requests when selecting next source

ajsutton:
## PR Description
During a recursive lookup, include nodes found by previous requests in the search in the list of sources to consider for the next round of requests.  This enables the actual recursive lookup behaviour, rather than only querying existing nodes we know about (and have previously pinged). As a result, we wind up gradually getting closer to the target node ID rather than gradually querying nodes further and further away (because we start from the closest we already know about and never request from the same node twice).

While the newly discovered nodes may not respond (we haven't seen a ping from them), if they do they can be immediately added to our buckets and if not the search will just move on to a different node.


rolfyone (approved):
LGTM

ajsutton:
Merged same effective change as part of #118

Issue #116, opened 2021-06-30
Don't include address in ENR until it is known


ajsutton:
### Description
Support starting with an ENR that contains no local IP information, then fill in the local address when it is determined from PONG responses.

Should still support specifying an initial IP to handle the case where a user explicitly sets the external IP.

PR #120, opened 2021-08-04
Give the NewAddressHandler more flexibilty in how it updates the node record

ajsutton:
## PR Description
Provide the old node record and new address to the `NewAddressHandler` instead of old and pre-updated node records. Update utility methods to add addresses to optionally support updating the TCP port as well as the address and UDP port.

This enables client code to delay adding the UDP or TCP address info to the ENR until the external IP address is known.

## Fixed Issue(s)
#116


rolfyone (approved):
LGTM

PR #121, opened 2021-08-12
Include active sessions in the stream of live nodes

ajsutton:
## PR Description
Modifies the `DiscoverySystem.streamLiveNodes()` method to include nodes we have an `AUTHENTICATED` session with, as well as node records that are stored in the k-buckets.

This increases the set of nodes a client can consider connecting to without having to increase the bucket storage size or maintain a separate unbounded node table as before. Active sessions are already actively maintained and limited in size, but may contain a lot more nodes than are in the k-buckets because they aren't sorted by distance.

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/DiscoveryManager.java
+++ b/src/main/java/org/ethereum/beacon/discovery/DiscoveryManager.java
@@ -7,6 +7,7 @@ package org.ethereum.beacon.discovery;
 import java.util.Collection;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.stream.Stream;
 import org.apache.tuweni.bytes.Bytes;
 import org.ethereum.beacon.discovery.schema.NodeRecord;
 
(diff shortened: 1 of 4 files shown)


rolfyone (approved):
LGTM

src/main/java/org/ethereum/beacon/discovery/DiscoveryManager.java
```diff
@@ -46,4 +47,6 @@
 
   /** Sends the TALKREQ so the specified {@code node} and returns the TALKRESP promise */
   CompletableFuture<Bytes> talk(NodeRecord nodeRecord, Bytes protocol, Bytes request);
+
+  Stream<NodeRecord> streamActiveSessions();
```
rolfyone, line 51: should be `@MustBeClosed`?

ajsutton: No need for that - it isn't from any resource that needs to be closed.

PR #129, opened 2021-12-13
Update dependencies

ajsutton:
## PR Description
Updates pretty much all the dependencies to latest version.


Nashatyrev (approved):
LGTM

src/test/java/org/ethereum/beacon/discovery/message/handler/EnrUpdateTrackerTest.java
```diff
@@ -63,7 +63,7 @@ void shouldNotRequestPeersEnrWhenPongSequenceNumberIsLower() {
     final NodeRecord currentNodeRecord =
         TestUtil.generateNode(8000)
             .getNodeRecord()
-            .withUpdatedCustomField("test", Bytes.EMPTY, Bytes.EMPTY);
+            .withUpdatedCustomField("test", Bytes.EMPTY, Bytes.of(1));
```
Nashatyrev, line 66: Just out of curiosity what was the reason of this change?

ajsutton: The bouncy castle update performs additional verification on public keys and it won't accept an empty byte array anymore.

PR #131, opened 2021-12-17
Optimization: Cache calculated NodeId in the NodeRecord

Nashatyrev:
## PR Description

`NodeId` calculation takes a lot of time and need to be cached 

изображение

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
+++ b/src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
@@ -5,6 +5,8 @@
 package org.ethereum.beacon.discovery.schema;
 
 import com.google.common.base.Preconditions;
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.Base64;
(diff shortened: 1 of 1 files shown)


tbenr (approved):
LGTM

ajsutton:
I think this is definitely good, I wonder if its worth also having a separate publicKey->nodeId cache (probably with some extra details around ensuring the same identity scheme is used etc).  There are a lot of node records we get from `NODES` messages which we parse and have to calculate the nodeId to validate they're from the requested distance. My guess is that a lot of these we've seen before and could just use the cached node ID even though it will be a new instance of `NodeRecord`.

ajsutton:
Going to merge this one as-is because it's such a quick easy win.  A separate cache of calculated node IDs is probably worth looking into as a follow up when time permits though.

Nashatyrev:
After this PR the signature verification is the major bottleneck (see flame graph below)  So it probably makes more sense to cache validated records (by their RLP hash?) assuming we are probably receiving many duplicates. 
(`NodeId` calculation takes roughly 5% of overall discovery CPU at the moment )

изображение

Nashatyrev:
We could actually combine both approaches into a single cache: 
```java
LRUMap<PublicKey, CacheEntry> nodeCache

class CacheEntry {
    final Bytes nodeId;
    Optional<Bytes32> lastVerifiedRLPHash;
}
```
I believe we could afford keeping up to 10K entries in that cache (would be roughly around 1Mb of heap at most)

PR #133, opened 2021-12-22
Move the caching of NodeId to be shared across NodeRecord instances

ajsutton:
## PR Description
The cache is now held in IdentitySchemaV4Interpreter, avoiding the need to create a new memoizing Supplier with every NodeRecord and avoiding the need to recalculate the NodeId if the same id is received from the network multiple times, which would result in multiple NodeRecord instances.


rolfyone (approved):
LGTM

Issue #125, opened 2021-11-11
Discovery: ConcurrentModificationException in Bucket.getLiveNodes


ajsutton:
### Description
```
java.util.concurrent.CompletionException: java.util.ConcurrentModificationException
	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320)
	at java.base/java.util.concurrent.CompletableFuture.uniHandle(CompletableFuture.java:936)
	at java.base/java.util.concurrent.CompletableFuture$UniHandle.tryFire(CompletableFuture.java:911)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2162)
	at java.base/java.util.concurrent.CompletableFuture$Timeout.run(CompletableFuture.java:2874)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList$ArrayListSpliterator.tryAdvance(ArrayList.java:1604)
	at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:129)
	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:527)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:513)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
[...]

PR #134, opened 2021-12-23
Make `KBucket` class thread safe

Nashatyrev:
## PR Description

Running large amount of discovery systems hit the `ConcurrentModificationException` with `KBucket.nodes` list: the list was simultaneously updated and iterated from different threads. 

As a solution just synchronized all public methods cause there are more than just `nodes` modifications. Seems to be deadlock-safe as the only external class is called under mutex: `LivenessChecker` which just asynchronously requests node ping.

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/storage/KBucket.java
+++ b/src/main/java/org/ethereum/beacon/discovery/storage/KBucket.java
@@ -41,22 +41,22 @@ class KBucket {
     this.clock = clock;
   }
 
-  public List<NodeRecord> getAllNodes() {
+  public synchronized List<NodeRecord> getAllNodes() {
     return nodes.stream().map(BucketEntry::getNode).collect(Collectors.toList());
   }
 
-  public List<NodeRecord> getLiveNodes() {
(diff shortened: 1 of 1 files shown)


ajsutton:
I'd have to dig more deeply to understand where the concurrent access is happening but I don't think this is the right answer.  `KBucket`should only be accessed from `KBuckets` (which appears to be the case) and it's `KBuckets` that should be providing the synchronisation. There's a fair few assumptions that `KBuckets` is providing the thread safety so I'd be very keen to properly understand how it's failing so we can make sure the fix is comprehensive.

ajsutton:
Ah based on the exception in #125 - I'm guessing that in `KBuckets` methods where we call `getBucket(distance).stream()`, even though it returned the `Optional` immediately, because it is converted to a stream, the `.flatMap` that actually accesses the bucket is executing after the method returns, and thus outside the `synchronized` block. I'll put up a PR which should fix it.

Nashatyrev:
Yep, the same exception here!
#135 is definitely the proper way to fix it 👍

PR #136, opened 2021-12-24
Create and use a single `SecureRandom` instance

Nashatyrev:
## PR Description

Creating `SecureRandom` every time is performance expensive ('eats' around 10% of overall discovery) and may potentially have security issues


ajsutton (approved):
LGTM.

PR #139, opened 2022-01-31
Remove direct dependency on vertx.

ajsutton:
## PR Description
vertx was only included as a direct dependency to override the default implementation of a tuweni-bytes method with a more optimised form. The latest tuweni has it's own `DelegatingBytes` class so switch to using that and remove ours along with the direct dependency on vertx.

Also delete the unused `Bytes52` class.


rolfyone (approved):
LGTM

PR #140, opened 2022-02-07
Validate ENR field keys ordering and uniqueness

Nashatyrev:
## PR Description

Special thanks to @asanso for finding and reporting this issue 👍 

Discovery doesn't validate field key ordering and uniqueness of wire ENRs. Thus any number of ENRs with valid signature could be generated and feed to the discovery. Any severe security impact was not found. 

As per [spec](https://github.com/ethereum/devp2p/blob/master/enr.md#record-structure): 
>The key/value pairs must be sorted by key and must be unique, i.e. any key may be present only once.


tbenr (approved):
LGTM

PR #141, opened 2022-02-09
Make `ExternalAddressSelector` pluggable

Nashatyrev:
## PR Description

Currently there is no mean to set a fixed external advertised address if doesn't match the address seen by other discv5 nodes: the local node record is overridden due to `Ping` responses. 

This PR makes `ExternalAddressSelector` pluggable so the client code could replace the default selector with `NOOP` or any other custom implementation

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/DiscoveryManagerImpl.java
+++ b/src/main/java/org/ethereum/beacon/discovery/DiscoveryManagerImpl.java
@@ -16,6 +16,7 @@ import org.apache.tuweni.bytes.Bytes;
 import org.ethereum.beacon.discovery.message.FindNodeMessage;
 import org.ethereum.beacon.discovery.message.PingMessage;
 import org.ethereum.beacon.discovery.message.TalkReqMessage;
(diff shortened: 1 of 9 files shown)


ajsutton (approved):
I don't think we specifically need this to have a fixed IP, but it seems useful to make the logic for selecting the new address pluggable anyway, so LGTM.

Nashatyrev:
> In Teku we enable/disable the automatic update of the ENR via the newAddressHandler [https://github.com/ConsenSys/teku/blob/6b5d530cf7cb00527aeaef5a472eab64f2011c78/ne[…]pegasys/teku/networking/p2p/discovery/discv5/DiscV5Service.java](https://github.com/ConsenSys/teku/blob/6b5d530cf7cb00527aeaef5a472eab64f2011c78/networking/p2p/src/main/java/tech/pegasys/teku/networking/p2p/discovery/discv5/DiscV5Service.java#L120)

Oh, right! That could indeed do the trick. 
Yep, let's anyway make this pluggable. Could be potentially of use later

PR #142, opened 2022-02-09
Misc updates: performance, logging, hacking

Nashatyrev:
## PR Description

A batch of minor updates: 

Performance:
- 3f2bba6f995099c346fe5bac2d2dd93df9c4a7d9: pass shared `SecureRandom` for new `NodeSession` instead of expensive creation a new instance
- 91446d7d45a80badaf3147ff25c6cb08fb522b64: use faster non-secure `Random` instead of `SecureRandom` to generate Envelope UUID which is used internally (for logging only)
- 6631150c5162805af8e0ab95248d6a4395803f31: use faster non-secure `Random` instead of `SecureRandom` to generate 'random message': there is no need for secure random here (even a static zeroes payload should not compromise the security) 

Logging: 
- c9ce3c53f35b1f3efe6603f75f2bfad21050e884: reduce logging exception noise when logging with `DEBUG` level

Hacking: 
- aa3396319a969eecca33cd294605685a5c069768: add an option to skip ENR signature verification for testing/hacking purposes

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/pipeline/Envelope.java
+++ b/src/main/java/org/ethereum/beacon/discovery/pipeline/Envelope.java
@@ -6,19 +6,20 @@ package org.ethereum.beacon.discovery.pipeline;
 
 import java.util.HashMap;
 import java.util.Map;
-import java.util.Random;
-import java.util.UUID;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicLong;
 
 /** Container for any kind of objects used in packet-messages-tasks flow */
 public class Envelope {
 
-  private final UUID id;
+  private static final AtomicLong ID_GENERATOR =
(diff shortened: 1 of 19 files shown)


src/main/java/org/ethereum/beacon/discovery/pipeline/Envelope.java
```diff
@@ -6,15 +6,19 @@
...
 /** Container for any kind of objects used in packet-messages-tasks flow */
 public class Envelope {
+
   private final UUID id;
   private final Map<Field<?>, Object> data = new HashMap<>();
 
   public Envelope() {
-    this.id = UUID.randomUUID();
+    Random random = ThreadLocalRandom.current();
+    this.id = new UUID(random.nextLong(), random.nextLong());
```
ajsutton, line 21: Given we only use this for logging, I wonder if we'd be better off just using `AtomicLong.getAndIncrement` to assign a simple unique identifier.  Could seed it with a large number if we wanted to keep it reasonably long and easier to search for.

ajsutton (approved):
LGTM.

PR #143, opened 2022-02-17
Switch to parsing RLP with tuweni-rlp instead of web3j

ajsutton:
## PR Description
Switches RLP parsing from webj3 to tuweni-rlp to avoid needing to parse into an intermediate representation before creating the internal types.


rolfyone (approved):
LGTM

PR #144, opened 2022-03-01
Update to gradle 7.4

ajsutton:
## PR Description
Updates to gradle 7.4 and adds a validation step that gradle wrapper jar is from an official release.


rolfyone (approved):
LGTM

PR #145, opened 2022-03-03
Add ignore list for nodes to avoid quick recheck after failed livenes…

### LINKED ISSUE #124: Check for recently evicted peers before adding them to the node table.
### Description
An enhancement when adding peers to node table.

### Discussion + Additional Information:

Let's say that a peer was dead (didn't reply to our last 10 pings).

We as a node evicted that particular peer.

And we are performing walks and when we do that, the evicted peer might still get added to our new table, If it is received from other lazy peer during FIND_NODES RPC.

And still we would use the same for our lookups etc. Added with exponential backoff, this might actually slow down disc?

Ping: @ajsutton. Does this make sense? or downright wrong?

zilm13:
…s check

## PR Description
Ignore table for peers, after peer fails to response on liveness check, it's added to ignore table for 30 seconds. During this period it's not added to the buckets and no rechecks are running for it.
I don't like the way I check this in `KBucket` but current architecture is weird and there is a big part of my miscount in it. I definitely want to see `SafeFuture` there one day.

## Fixed Issue(s)

Should fix #124

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/database/ExpirationSet.java
+++ b/src/main/java/org/ethereum/beacon/discovery/database/ExpirationSet.java
@@ -12,7 +12,6 @@ import java.util.HashSet;
 import java.util.Set;
 import java.util.TreeSet;
 import kotlin.Pair;
-import org.jetbrains.annotations.NotNull;
 
 /**
  * Set-alike collection with data expiring in configured time. This structure is not thread safe,
@@ -104,7 +103,7 @@ public class ExpirationSet<V extends Comparable<V>> {
     return true;
   }
 
-  public boolean addAll(@NotNull Collection<? extends V> c) {
+  public boolean addAll(Collection<? extends V> c) {
     clearExpired();
     c.forEach(this::add);
     return true;
--- a/src/main/java/org/ethereum/beacon/discovery/database/ExpirationSet.java
+++ b/src/main/java/org/ethereum/beacon/discovery/database/ExpirationSet.java
@@ -7,34 +7,30 @@ package org.ethereum.beacon.discovery.database;
 import java.time.Clock;
 import java.time.Duration;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.Set;
 import java.util.TreeSet;
 import kotlin.Pair;
 import org.jetbrains.annotations.NotNull;
 
 /**
- * Set with data expiring in configured time
+ * Set-alike collection with data expiring in configured time. This structure is not thread safe,
+ * please, synchronize usage
  *
- * @param <V> data type
+ * @param <V> data type should extend Comparable<V> to avoid collisions
  */
-public class ExpirationSet<V> implements Set<V> {
+public class ExpirationSet<V extends Comparable<V>> {
 
   private final long expirationDelayMillis;
   private final Clock clock;
   private final long size;
-
-  TreeSet<Pair<Long, V>> dataExpiration =
+  private final TreeSet<Pair<Long, V>> dataExpiration =
       new TreeSet<>(
-          (o1, o2) -> {
-            int expirationDeltaSignum = Long.signum(o1.getFirst() - o2.getFirst());
-            if (expirationDeltaSignum != 0) {
-              return expirationDeltaSignum;
-            }
-            return Integer.compare(o1.getSecond().hashCode(), o2.getSecond().hashCode());
-          });
-  Set<V> data = new HashSet<>();
+          Comparator.comparingLong(Pair<Long, V>::getFirst)
+              .reversed()
+              .thenComparing(Pair::getSecond));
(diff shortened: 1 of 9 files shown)


src/main/java/org/ethereum/beacon/discovery/database/ExpirationSet.java
```diff
@@ -0,0 +1,167 @@
...
+  TreeSet<Pair<Long, V>> dataExpiration =
+      new TreeSet<>(
+          (o1, o2) -> {
+            int expirationDeltaSignum = Long.signum(o1.getFirst() - o2.getFirst());
+            if (expirationDeltaSignum != 0) {
+              return expirationDeltaSignum;
+            }
+            return Integer.compare(o1.getSecond().hashCode(), o2.getSecond().hashCode());
+          });
+  Set<V> data = new HashSet<>();
```
ajsutton, line 37: These can both be `private final`.

src/main/java/org/ethereum/beacon/discovery/database/ExpirationSet.java
```diff
@@ -0,0 +1,167 @@
...
+  private final long size;
+
+  TreeSet<Pair<Long, V>> dataExpiration =
+      new TreeSet<>(
+          (o1, o2) -> {
+            int expirationDeltaSignum = Long.signum(o1.getFirst() - o2.getFirst());
+            if (expirationDeltaSignum != 0) {
+              return expirationDeltaSignum;
+            }
+            return Integer.compare(o1.getSecond().hashCode(), o2.getSecond().hashCode());
```
ajsutton, line 35: This comparison isn't safe as `hashCode` can return the same value for different objects. I suspect we should rework this to just store the node id which is a `Bytes` and implements `Comparable` itself.  Then the Compartor could be created with:
```
new TreeSet<>(
          Comparator.comparingLong(Pair<Long, Bytes>::getFirst)
              .reversed()
              .thenComparing(Pair::getSecond));
```

Though you should double check if the `.reversed()` should be there or not. :)

src/main/java/org/ethereum/beacon/discovery/database/ExpirationSet.java
```diff
@@ -0,0 +1,167 @@
...
+
+  @Override
+  public boolean retainAll(@NotNull Collection<?> c) {
+    throw new RuntimeException("Not implemented yet");
+  }
+
+  @Override
+  public boolean removeAll(@NotNull Collection<?> c) {
+    throw new RuntimeException("Not implemented yet");
+  }
```
ajsutton, line 160: Rather than leaving methods unimplemented, I'd just not implement the generic `Set` interface as we don't need a generic data structure here, just the `add`, and `contains` methods with the automatic expiry.

src/main/java/org/ethereum/beacon/discovery/liveness/LivenessChecker.java
```diff
@@ -54,6 +64,7 @@ private void sendPing(final NodeRecord node) {
             (__, error) -> {
               if (error != null) {
                 LOG.trace("Liveness check failed for node {}", node, error);
+                ignoredNodes.add(node);
               }
```
ajsutton, line 68: We should move this into the `synchronized` block below and then `ignoreNodes` won't need to be thread safe.

src/main/java/org/ethereum/beacon/discovery/liveness/LivenessChecker.java
```diff
@@ -3,22 +3,32 @@
...
+import org.ethereum.beacon.discovery.database.ExpirationSet;
 import org.ethereum.beacon.discovery.schema.NodeRecord;
 
 public class LivenessChecker {
   private static final Logger LOG = LogManager.getLogger();
 
   static final int MAX_CONCURRENT_PINGS = 3;
   static final int MAX_QUEUE_SIZE = 1000;
+  static final int MAX_IGNORE_SET_SIZE = 1000;
+  static final Duration ignoreDuration = Duration.ofSeconds(30);
```
ajsutton, line 23: nit: This is a constant so should be named `IGNORE_DURATION`.

ajsutton (commented):
Basic structure looks good - mostly just the concurrency issues to sort out.

src/main/java/org/ethereum/beacon/discovery/database/ExpirationSet.java
zilm13: Yeah, actually there could be disintegration  between `data` and `dataExpiration` in rare case and this will cause whole structure to be buggy

src/main/java/org/ethereum/beacon/discovery/liveness/LivenessChecker.java
zilm13: yeah, was careless

zilm13:
@ajsutton updated according to feedback

ajsutton (approved):
LGTM.

PR #148, opened 2022-06-03
Remove usage of web3j library

### LINKED ISSUE #147: Replace web3j crypto
Replace the web3j crypto functions with something more reliable.  Possibly using BouncyCastle directly or maybe Tuweni crypto. In particular the ECDSA functions.

zilm13:
## PR Description
Removes `web3j-crypto` and `web3j-rlp` usage across code in favor of `tuweni`

Design consideration:

There is a `SecretKey` in `tuweni-crypto` with `Bytes32` source of the private key which is not copied until you use some guarded `Bytes` source, this `SecretKey` implements [finalize](https://github.com/apache/incubator-tuweni/blob/main/crypto/src/main/java/org/apache/tuweni/crypto/SECP256K1.java#L445-L457) filling source with `00`'s, and when you create another `SecretKey` with the same `Bytes32` source and second `SecretKey` marked to be collected with GC, private key is erased in both as both uses the same instance of `Bytes32` (it was not actually copied). From one side it's a good thing to guarantee erasure of data but it forces to use single `SecretKey` instance all over the library avoiding to create it when really needed.

I've considered wrapping `SecretKey` with some `Signer` interface to hide implementation and benefit from test's `NoopSigner`, but when parameterized it goes straight from `IdentitySchemaInterpreter` to `NodeRecord` which is used almost in every class in the package and should take more time to implement while I already spent more time than I should on this task. 
## Fixed Issue(s)

Fixes #147


src/main/java/org/ethereum/beacon/discovery/util/Functions.java
```diff
@@ -15,153 +12,154 @@
...
-  private static boolean SKIP_SIGNATURE_VERIFY = false;
+  static {
+    Security.addProvider(PROVIDER = new BouncyCastleProvider());
+  }
 
   /** SHA2 (SHA256) */
-  public static Bytes hash(Bytes value) {
-    return Hashes.sha256(value);
+  public static Bytes32 hash(final Bytes value) {
+    return Bytes32.wrap(Hashes.sha256(value));
```
ajsutton, line 56: Seems like we could just make `Hashes.sha256` return a `Bytes32`.

src/test/java/org/ethereum/beacon/discovery/schema/NodeSessionTest.java
```diff
@@ -156,4 +158,19 @@ void createNextRequest_shouldNotResetAuthenticatedStatesWhenRequestTimesOut() {
     timeoutHandler.run();
     assertThat(session.getState()).isEqualTo(SessionState.AUTHENTICATED);
   }
+
+  private Request<?> createRequestMock() {
+    final Request<?> request = mock(Request.class);
+    when(request.getResultPromise()).thenReturn(new CompletableFuture<>());
+    // avoid falling when logging is on
```
ajsutton, line 165:
```suggestion
    // avoid failing when logging is on
```

src/main/java/org/ethereum/beacon/discovery/util/Functions.java
```diff
@@ -15,153 +12,154 @@
...
   private static final int RECIPIENT_KEY_LENGTH = 16;
   private static final int INITIATOR_KEY_LENGTH = 16;
   private static final int AUTH_RESP_KEY_LENGTH = 16;
   private static final int MS_IN_SECOND = 1000;
 
   private static final Supplier<SecureRandom> SECURE_RANDOM = Suppliers.memoize(SecureRandom::new);
 
-  private static boolean SKIP_SIGNATURE_VERIFY = false;
+  static {
+    Security.addProvider(PROVIDER = new BouncyCastleProvider());
```
ajsutton, line 51: It would be better to init `PROVIDER` inline and then just have 
```suggestion
    Security.addProvider(PROVIDER);
```

That said, I don't see where `PROVIDER` is being used.  Why did this move from `Hashes` to here?

src/main/java/org/ethereum/beacon/discovery/util/Functions.java
```diff
@@ -15,153 +12,154 @@
...
-    return Hashes.sha256(value);
+  public static Bytes32 hash(final Bytes value) {
+    return Bytes32.wrap(Hashes.sha256(value));
   }
 
   /** SHA3 (Keccak256) */
-  public static Bytes hashKeccak(Bytes value) {
-    return Bytes.wrap(Hash.sha3(value.toArray()));
+  public static Bytes32 hashKeccak(final Bytes value) {
+    return Bytes32.wrap(Hash.keccak256(value.toArray()));
```
ajsutton, line 61: Wouldn't the below be enough?

```suggestion
    return Hash.keccak256(value);
```

src/main/java/org/ethereum/beacon/discovery/util/Functions.java
```diff
@@ -15,153 +12,154 @@
...
-            extractBytesFromUnsignedBigInt(calculatedPubKey, PUBKEY_SIZE))) {
-          return true;
-        }
+      final boolean verifyV1 =
+          SECP256K1.verifyHashed(
+              hashedMessage,
+              Signature.create(
+                  (byte) 1,
+                  signature.slice(0, 32).toUnsignedBigInteger(),
+                  signature.slice(32).toUnsignedBigInteger()),
```
ajsutton, line 96: We should assert the signature has the expected length to prevent getting weird errors from these slice calls.

src/main/java/org/ethereum/beacon/discovery/util/Functions.java
```diff
@@ -15,153 +12,154 @@
...
+        return true;
       }
-    } catch (final IllegalArgumentException e) {
+      return SECP256K1.verifyHashed(
+          hashedMessage,
+          Signature.create(
+              (byte) 0,
+              signature.slice(0, 32).toUnsignedBigInteger(),
+              signature.slice(32).toUnsignedBigInteger()),
+          publicKey);
```
ajsutton, line 107: It would be good to deduplicate these two checks. The only difference should be the `v` value so could either extract a method that takes pubKey, signature, message and v value or just use a `for (byte v = 0; v <= 1; v++)` loop.

ajsutton (commented):
Looking good. Just a few minor things and a couple of queries.

ajsutton (approved):
LGTM. Nice work.

PR #150, opened 2022-10-19
Ensure no unexpected trailing data for all message types

jtraglia:
## PR Description

Noticed that `FindNodeMessage` called `RlpUtil#checkComplete` but the other message types did not.

* Call `RlpUtil.checkComplete(reader)` after reading all expected values.
* For consistency:
  * Rename `listReader` -> `reader` in `FindNodeMessage`.
  * Remove number (`1`) from end of variable names.

PR #155, opened 2022-10-20
Add shouldRejectTrailingBytes test for messages

jtraglia:
## PR Description

So I messed up in #150. I didn't notice that `RlpUtil#readRlpList` calls `checkComplete`. Sorry about that.

https://github.com/ConsenSys/discovery/blob/9bb8ad3942f81450fe54a78c66f0b61e58c45b0b/src/main/java/org/ethereum/beacon/discovery/util/RlpUtil.java#L39-L47

This PR will:

* Remove extraneous `checkComplete` calls.
* Add `assertRoundTrip` and `assertRejectTrailingBytes` to `TestUtil`.
* For each message type test class:
  * Make the decoder object static final (and uppercase its name).
  * For reusability, extract "original" message into a static final field.
  * Add new `shouldRejectTrailingBytes` test.
  * Replace body of `shouldRoundTrip` with `assertRoundTrip`.

Also, unrelated to the core part of this PR:

* Fix minor format issue (`TAUthData` -> `TAuthData`) in `HeaderImpl`.
* Fix a couple of `it's` vs `its` in comments.


ajsutton (approved):
LGTM. Definitely good to add these missed tests.

PR #156, opened 2022-10-20
Ensure pong response recipient port is a uint16

jtraglia:
## PR Description

According to the spec, `recipient-port` is a (unsigned) 16-bit integer. But the implementation of `PongMessage` allowed any 32-bit integer to be used. I've added checks (`isPortValid` from Teku) to ensure its compliance.

* https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#pong-response-0x02

We could add checks to the constructor, but I don't think that's necessary. Maybe in a different PR.


ajsutton (approved):
LGTM.

PR #157, opened 2022-10-20
Throw decoding exception if there's an invalid node distance

jtraglia:
## PR Description

According to the [spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md#node-table), there are 256 buckets and 0 is reserved for our node, so a distance value can be in `[0, 256]`.

> For each `0 ≤ i < 256`, every node keeps a k-bucket for nodes of `logdistance(self, n) == I`. The Node Discovery Protocol uses `k = 16`, i.e. every k-bucket contains up to 16 node entries.

I see that this is enforced in the `KBuckets` class, but I think we should throw a decoding error if we receive a `FindNodeMessage` with an invalid distance. We could just ignore that distance, but I think we should ignore the message:

https://github.com/ConsenSys/discovery/blob/d780ba4642db8acd1a6a8253d9e86018eafb8b89/src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java#L22-L29

Also, in my opinion, the spec is pretty vague about this detail. My first impression told me that a distance of 256 is invalid, but after I thought more about it, it seems valid. I'm still kind of confused though 🤷

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/message/FindNodeMessage.java
+++ b/src/main/java/org/ethereum/beacon/discovery/message/FindNodeMessage.java
@@ -14,6 +14,7 @@ import org.apache.tuweni.rlp.RLP;
 import org.apache.tuweni.rlp.RLPReader;
 import org.apache.tuweni.rlp.RLPWriter;
 import org.ethereum.beacon.discovery.util.DecodeException;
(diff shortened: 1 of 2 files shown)


ajsutton (approved):
LGTM. I think you're right that 256 is a valid distance (and also agree that the spec is unclear...)

PR #158, opened 2022-10-24
Update errorprone & do ConsenSys errorprone checks

jtraglia:
## PR Description

* Update errorprone to the latest version (`2.16`)
  * Btw it's a little annoying that `2.16.0` isn't valid yet.
* Update the errorprone gradle plugin to the latest version.
* Do checks from [`errorprone-checks`](https://github.com/ConsenSys/errorprone-checks) (by ConsenSys).
* Like Teku, suppress the one instance of `DoNotInvokeMessageDigestDirectly`.
* Like Teku, enable a variety of experimental checks that I think are good.
* Remove disabled checks for Java 12 that were broken.
  * I don't think this is an issue any more.
* Fix findings from `MissingBraces` and `UnnecessarilyFullyQualified`.
  * Will fix findings from `JavaCase` in a different PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ plugins {
   id 'com.github.ben-manes.versions' version '0.41.0'
   id 'com.github.hierynomus.license' version '0.16.1'
   id 'io.spring.dependency-management' version '1.0.11.RELEASE'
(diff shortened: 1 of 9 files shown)


ajsutton (approved):
LGTM.

PR #159, opened 2022-10-25
Check for more required fields

jtraglia:
## PR Description

So for these handlers, I believe these fields should be required. As a rule of thumb, if `envelope.get(FIELD)` is called in the handler, `FIELD` should be required, right? For one instance, `MessageHandler`, all of the specific message handlers (`NodesHandler`, `PingHandler`, etc) seem to assume a valid session.

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/pipeline/handler/MessageHandler.java
+++ b/src/main/java/org/ethereum/beacon/discovery/pipeline/handler/MessageHandler.java
@@ -40,9 +40,6 @@ public class MessageHandler implements EnvelopeHandler {
     if (!HandlerUtil.requireField(Field.MESSAGE, envelope)) {
       return;
     }
-    if (!HandlerUtil.requireField(Field.SESSION, envelope)) {
-      return;
-    }
     if (!HandlerUtil.requireSessionWithNodeRecord(envelope)) {
       return;
     }
--- a/src/main/java/org/ethereum/beacon/discovery/pipeline/handler/MessageHandler.java
+++ b/src/main/java/org/ethereum/beacon/discovery/pipeline/handler/MessageHandler.java
@@ -40,6 +40,9 @@ public class MessageHandler implements EnvelopeHandler {
     if (!HandlerUtil.requireField(Field.MESSAGE, envelope)) {
       return;
     }
+    if (!HandlerUtil.requireField(Field.SESSION, envelope)) {
+      return;
+    }
     if (!HandlerUtil.requireNodeRecord(envelope)) {
       return;
     }
--- a/src/main/java/org/ethereum/beacon/discovery/pipeline/handler/MessageHandler.java
+++ b/src/main/java/org/ethereum/beacon/discovery/pipeline/handler/MessageHandler.java
@@ -43,7 +43,7 @@ public class MessageHandler implements EnvelopeHandler {
     if (!HandlerUtil.requireField(Field.SESSION, envelope)) {
       return;
     }
-    if (!HandlerUtil.requireNodeRecord(envelope)) {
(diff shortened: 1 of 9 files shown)


ajsutton:
Will have to dig into this one more deeply. The risk is that we wind up silently ignoring an event which makes it very hard to debug what went wrong. Currently we'd throw an exception that would get logged so would see that something is wrong.

I really dislike the whole pipeline and filtering pattern in use though so need to really think through how this fits together.

src/main/java/org/ethereum/beacon/discovery/pipeline/handler/NextTaskHandler.java
```diff
@@ -47,6 +47,9 @@ public static void tryToSendAwaitTaskIfAny(
 
   @Override
   public void handle(Envelope envelope) {
+    if (!HandlerUtil.requireField(Field.SESSION, envelope)) {
+      return;
+    }
```
ajsutton, line 52: Also covered by `requiredNodeRecord`

src/main/java/org/ethereum/beacon/discovery/pipeline/handler/MessageHandler.java
```diff
@@ -40,6 +40,9 @@ public void handle(Envelope envelope) {
     if (!HandlerUtil.requireField(Field.MESSAGE, envelope)) {
       return;
     }
+    if (!HandlerUtil.requireField(Field.SESSION, envelope)) {
+      return;
+    }
     if (!HandlerUtil.requireNodeRecord(envelope)) {
```
ajsutton, line 46: This is covered by the `HandlerUtil.requireNodeRecord` below.  Could possibly rename that to `requireSessionWithNodeRecord` for clarity though.

src/main/java/org/ethereum/beacon/discovery/pipeline/handler/WhoAreYouPacketHandler.java
```diff
@@ -50,6 +50,12 @@ public void handle(final Envelope envelope) {
     if (!HandlerUtil.requireField(Field.PACKET_WHOAREYOU, envelope)) {
       return;
     }
+    if (!HandlerUtil.requireField(Field.MASKING_IV, envelope)) {
+      return;
+    }
+    if (!HandlerUtil.requireField(Field.SESSION, envelope)) {
+      return;
+    }
```
ajsutton, line 58: Covered by a `requireNodeRecord` call.

src/main/java/org/ethereum/beacon/discovery/pipeline/handler/WhoAreYouPacketHandler.java
```diff
@@ -50,6 +50,12 @@ public void handle(final Envelope envelope) {
     if (!HandlerUtil.requireField(Field.PACKET_WHOAREYOU, envelope)) {
       return;
     }
+    if (!HandlerUtil.requireField(Field.MASKING_IV, envelope)) {
```
ajsutton, line 53: This is an interesting one - we do explicitly check for it later and throw a very explicit exception if it's missing (around line 85).

I'd be tempted to move that check up higher so it happens before we start performing any actual work but keep it throwing an exception rather than just ignoring the packet silently as an unmet requirement.

src/main/java/org/ethereum/beacon/discovery/pipeline/handler/MessageHandler.java
jtraglia: Ah, you're right. I missed that. Thanks, will fix for all instances 👍

ajsutton (approved):
LGTM. I just need to work out why circleci didn't run...

PR #160, opened 2022-10-25
Add shouldRejectInvalidTotalPackets test

jtraglia:
## PR Description

Well I misinterpreted [the spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#nodes-response-0x04) and thought the following sentences meant that `FindNodeResponseHandler` should ignore excess packets, instead of rejecting the response, after the allowed maximum for total.

> Implementations may place a limit on the allowed maximum for total. If exceeded, additional responses may be ignored. 

I started to "fix" it but then I noticed that it said "additional responses" and not "additional packets". So I decided to add some tests to ensure that functionality works as expected. Also, updated two things in the handler that I thought helped readability. Should we ignore additional responses after a bad response?


ajsutton (approved):
LGTM. I think it would make sense to ignore all packets after one is invalid.

PR #161, opened 2022-10-26
Make streamClosestNodes synchronized

jtraglia:
## PR Description

Not 100% confident, but I think this method should be synchronized. Every other public method in this class is synchronized; this method was added after the others. `this` passed to `KBucketsIterator` could change during iteration.

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java
+++ b/src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java
@@ -80,7 +80,7 @@ public class KBuckets {
     }
   }
 
-  public Stream<NodeRecord> streamClosestNodes(Bytes nodeId) {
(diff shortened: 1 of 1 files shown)


ajsutton:
`streamClosestNodes` doesn't need to be synchronised because it's not accessing any of the mutable fields in `KBuckets`.  As you note it does pass the `KBuckets` instance into the `KBucketsIterator` instance it creates but that's just the public reference to the instance. Every method `KBucketsIterator` calls on the `KBuckets` instance it has are synchronised which ensures the access is thread safe.

Worth nothing though that `KBucketsIterator` (and thus the returned `Stream`) doesn't attempt to iterate over a fixed snapshot of the state of `KBuckets`. Changes made after the creation of the stream may or may not be reflected in the stream. That's the same iteration behaviour as things like `ConcurrentHashMap`.

jtraglia:
Ah, yes, that makes sense. Thank for you explaining it!

PR #162, opened 2022-10-26
Simplify logDistance function

jtraglia:
## PR Description

I think this version is...

*  easier to understand,
* matches the description in the method comment,
* and is most likely faster.


ajsutton (approved):
LGTM. Definitely simpler and easier to read.

PR #163, opened 2022-10-27
Fix JavaCase findings

jtraglia:
## PR Description

This touches a lot of files, but the changes are pretty minor. A lot fewer findings than Teku.

* Enable the `JavaCase` errorprone check.
* Change severity from INFO to WARNING.
* Rename `logger` (they're all static final) to `LOG` (simpler than `LOGGER`).
* Rename all static final variables from camelCase to UPPER_SNAKE.
* Rename local variables from UPPER_SNAKE to camelCase (there were just a few).
* Rename `Functions.hkdf_expand` to `Functions.hkdfExpand`.
* Add `static` qualifier to a few variables that should probably have it.
* Suppress warnings for `DiscoveryManagerTest` because I think it's easier to read how it is currently.


ajsutton (approved):
LGTM. Thank you as always. :)

PR #164, opened 2022-10-31
Query distances from left/right of target distance

jtraglia:
## PR Description

So the [lookup section of the spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-theory.md#lookup) isn't super clear in this regard, but I think `findNodes` isn't querying the ideal distances. Given a target distance of 10, it will query 10, 11, and 12. I think it should query 9, 10, 11. 9 is closer to 10 than 12. Slight distinction but it may help. Also, while I don't think it's common, the previous implementation doesn't work very well with a target distance of 255 and 256; it will only query one or two distances, instead of the typical three. Thoughts?

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java
+++ b/src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java
@@ -23,7 +23,7 @@ public class KBuckets {
    * Minimum distance we create a bucket for. 0 is our local node record and negative distances
    * aren't allowed.
    */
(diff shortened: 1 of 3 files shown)


jtraglia:
I just realized that the previous implementation would query 4 distances, instead of 3 like I thought. I'll wait for your initial review before changing `NUM_DISTANCES_TO_QUERY` to 4, assuming that's what you want.

jtraglia:
Thanks for the detailed response. Makes sense to me. I'll close this.

PR #165, opened 2022-11-02
Lookup lower distances if there's space

jtraglia:
## PR Description

Alternative version of #164 based on [this comment](https://github.com/ConsenSys/discovery/pull/164#issuecomment-1297698736). We want to prioritize higher distances, but...

> It would be reasonable to add lower distances if we've reached MAX_DISTANCE though.

This PR does exactly that.

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/task/DiscoveryTaskManager.java
+++ b/src/main/java/org/ethereum/beacon/discovery/task/DiscoveryTaskManager.java
@@ -112,14 +112,19 @@ public class DiscoveryTaskManager {
 
     // We want to prioritize higher distances in the request over lower ones
     // because lower distances are less likely to contain nodes.
-    for (int offset = 1; distances.size() < LOOKUP_REQUEST_LIMIT; offset++) {
-      if (targetDistance + offset <= KBuckets.MAXIMUM_BUCKET) {
-        distances.add(targetDistance + offset);
-        continue;
+    for (int offset = 1; targetDistance + offset <= KBuckets.MAXIMUM_BUCKET; offset++) {
+      if (distances.size() >= LOOKUP_REQUEST_LIMIT) {
+        break;
       }
-      if (targetDistance - offset >= KBuckets.MINIMUM_BUCKET) {
-        distances.add(targetDistance - offset);
+      distances.add(targetDistance + offset);
+    }
+
+    // If there's any space left, add lower distances.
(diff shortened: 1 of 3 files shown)


src/main/java/org/ethereum/beacon/discovery/task/DiscoveryTaskManager.java
```diff
@@ -96,14 +100,38 @@ private CompletableFuture<Collection<NodeRecord>> performSearchForNewPeers() {
...
-        distance++) {
-      distances.add(distance);
+    distances.add(targetDistance);
+
+    // We want to prioritize higher distances in the request over lower ones
+    // because lower distances are less likely to contain nodes.
+    for (int offset = 1; targetDistance + offset <= KBuckets.MAXIMUM_BUCKET; offset++) {
+      if (distances.size() >= LOOKUP_REQUEST_LIMIT) {
+        break;
+      }
```
ajsutton, line 118: nit: I think I'd simplify this to:
```suggestion
    for (int offset = 1; targetDistance + offset <= KBuckets.MAXIMUM_BUCKET && distances.size() < LOOKUP_REQUEST_LIMIT; offset++) {
```

src/main/java/org/ethereum/beacon/discovery/task/DiscoveryTaskManager.java
```diff
@@ -96,14 +100,38 @@ private CompletableFuture<Collection<NodeRecord>> performSearchForNewPeers() {
...
+      distances.add(targetDistance + offset);
+    }
+
+    // If there's any space left, add lower distances.
+    for (int offset = 1; targetDistance - offset >= KBuckets.MINIMUM_BUCKET; offset++) {
+      if (distances.size() >= LOOKUP_REQUEST_LIMIT) {
+        break;
+      }
```
ajsutton, line 126: Similar simplification here.

ajsutton (commented):
LGTM just a couple of minor simplifications I'd suggest.

src/main/java/org/ethereum/beacon/discovery/task/DiscoveryTaskManager.java
jtraglia: Good call. I've updated those.

ajsutton (approved):
LGTM. Thanks.

PR #168, opened 2022-11-04
Use Guava cache for ExpirationSet to simplify and optimise implementation

ajsutton:
## PR Description
Simplify and optimise `ExpirationSet` by using guava cache. Fixes a bug where the newest item was removed when the set was full rather than the oldest.
Optimise stats collection in KBucket to avoid creating new lists of all nodes.

## Fixed Issue(s)
https://github.com/ConsenSys/teku/issues/6371

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/database/ExpirationSet.java
+++ b/src/main/java/org/ethereum/beacon/discovery/database/ExpirationSet.java
@@ -4,14 +4,15 @@
 
 package org.ethereum.beacon.discovery.database;
 
+import static com.google.common.base.Preconditions.checkArgument;
+
+import com.google.common.base.Ticker;
+import com.google.common.cache.CacheBuilder;
 import java.time.Clock;
 import java.time.Duration;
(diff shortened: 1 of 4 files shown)


src/main/java/org/ethereum/beacon/discovery/database/ExpirationSet.java
```diff
@@ -21,96 +22,53 @@
...
    */
-  public ExpirationSet(final long expirationDelayMillis, final Clock clock, final long size) {
-    this.expirationDelayMillis = expirationDelayMillis;
-    this.clock = clock;
-    if (size < 1) {
-      throw new RuntimeException("Minimal size is 1");
-    }
-    this.size = size;
+  public ExpirationSet(final long expirationDelayMillis, final Clock clock, final long maxSize) {
+    checkArgument(maxSize > 0, "Minimal size is 1");
```
rolfyone, line 35: minimum size?

rolfyone (approved):
LGTM just a nit

PR #170, opened 2022-11-15
Remove old/unused database classes

jtraglia:
## PR Description

It's my understanding that these files were used in the original implementation, but are not longer needed.

* `DataSource` was used for the node table.
* `HoleyList` was used for the node index table.
* `SingleValueSource` was used for the home node.

If these aren't being used, I think it's a good idea to remove them.

I'm leaving `ExpirationSet` because it is used. But I think it could be moved to the `storage` directory.


jtraglia:
Not sure what's up with CI. I think it needs to be kicked.

ajsutton (approved):
LGTM. Definitely worth clearing out dead code.

PR #171, opened 2022-12-18
Implement source address filtering

ajsutton:
## PR Description
Implements an address access policy to allow library users to restrict which IP addresses are allowed for incoming and outgoing packets as well as for accepted node records.

## Fixed Issue(s)
https://github.com/ConsenSys/teku/issues/6242


src/main/java/org/ethereum/beacon/discovery/pipeline/handler/HandshakeMessagePacketHandler.java
```diff
@@ -130,7 +134,7 @@ public void handle(Envelope envelope) {
 
       session.setState(AUTHENTICATED);
       envelope.remove(Field.PACKET_HANDSHAKE);
-      enr.ifPresent(session::onNodeRecordReceived);
+      enr.filter(addressAccessPolicy::allow).ifPresent(session::onNodeRecordReceived);
```
tbenr, line 137: I'm wondering if we should fail the handshake completely, so apply the filter as early as possible..

ajsutton: Yeah seems reasonable.

tbenr (approved):
LGTM

PR #172, opened 2023-08-03
Bouncy Castle to 1.76 using the Java 8 variant

fab-10:
## PR Description

Upgrading Bouncy Castle to the latest version and switching to the variant that requires at least Java 8.

Does anyone knows if there is risk in dropping per Java 8 support?


StefanBratanov (approved):
Looks good. In any case, we set source and target compatibility to 11, so we don't support 8, so no worries there.

fab-10:
@StefanBratanov could you help with the publishing failure on main?

StefanBratanov:
> @StefanBratanov could you help with the publishing failure on main?

Yeah, I am looking into it. Looks like it has been failing for a while.

StefanBratanov:
> > @StefanBratanov could you help with the publishing failure on main?
> 
> Yeah, I am looking into it. Looks like it has been failing for a while.

@fab-10 Done. Publishing on last commit looks good.

Issue #176, opened 2024-05-24
Support listening and advertising dual-stack (IPv4 and IPv6)


StefanBratanov:
**Description**

In order to support dual-stack (IPv4 and IPv6), discovery should be able to configure multiple listening addresses and advertising dual-stack ENR.

PR #180, opened 2024-06-25
IPv4/IPv6 additional handling + dependencies updates

StefanBratanov:
## PR Description

- Upgrade ilbraries + gradle
- Align usage of `NewAddressHandler` to be compatible when using both IPv4 and IPv6

## Fixed Issue(s)
related to #176


StefanBratanov:
> LGTM nit: Just a bit uncomfortable that `switch`es handling IPv4/IPv6 cases are spreaded across classes. Any chances we could group them in a single place?

Yeah, I agree about the switches handling. Added a commit to simplify using ternary operators. I think I am more likely to raise an issue to think about way to handle all those IPv4/IPv6 handling better. It will require some thought.

PR #183, opened 2024-10-14
Update tuweni library.

david-ry4n:
## PR Description
Update the tuweni library.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,10 +26,11 @@ repositories {
 }
 
 dependencies {
-  implementation 'org.apache.tuweni:tuweni-bytes'
-  implementation 'org.apache.tuweni:tuweni-crypto'
-  implementation 'org.apache.tuweni:tuweni-rlp'
-  implementation 'org.apache.tuweni:tuweni-units'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -26,10 +26,11 @@ repositories {
 }
 
 dependencies {
-  implementation 'org.apache.tuweni:tuweni-bytes'
-  implementation 'org.apache.tuweni:tuweni-crypto'
-  implementation 'org.apache.tuweni:tuweni-rlp'
-  implementation 'org.apache.tuweni:tuweni-units'
+  implementation 'io.vertx:vertx-core'
```
StefanBratanov, line 29: What is the need of vertx-core dependency?

david-ry4n: It's a dependency for tuweni. Building locally, it would only compile if I added the dependency. For some reason tuweni wasn't bringing in that dependency automatically.

StefanBratanov: sure, sounds good, we bring in that dependency in Teku anyways

StefanBratanov (approved):
LGTM

PR #185, opened 2025-01-28
Ensure all node records are valid length

Matilda-Clerke:
## PR Description
Add a check that the ENR is no more than 300 bytes, per the [spec](https://eips.ethereum.org/EIPS/eip-778#rlp-encoding) before attempting to parse
## Fixed Issue(s)
https://github.com/Consensys/protocol-misc/issues/1022


src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
```diff
@@ -47,36 +47,44 @@ public class NodeRecord {
...
-  private NodeRecord(IdentitySchemaInterpreter identitySchemaInterpreter, UInt64 seq) {
-    this.seq = seq;
-    this.signature = MutableBytes.create(96);
-    this.identitySchemaInterpreter = identitySchemaInterpreter;
+    this.fields.putAll(fields);
+    // serialise to check size
+    Bytes serializedNodeRecord = serialize();
+    checkArgument(
+        serializedNodeRecord.size() <= MAX_ENCODED_SIZE,
+        "Node record exceeds maximum encoded size");
```
Matilda-Clerke, line 62: Did a little refactoring here to ensure the NodeRecord is fully constructed after the constructor is run. Previously, the "fields" were added after

src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
```diff
@@ -47,36 +47,44 @@ public class NodeRecord {
...
-    rawFields.forEach(
-        (key, value) -> nodeRecord.set(key, ENR_FIELD_INTERPRETER.decode(key, value)));
-    return nodeRecord;
+    return new NodeRecord(
+        identitySchemaInterpreter,
+        seq,
+        signature,
+        rawFields.entrySet().stream()
+            .map((e) -> Map.entry(e.getKey(), ENR_FIELD_INTERPRETER.decode(e.getKey(), e.getValue())))
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
```
Matilda-Clerke, line 87: More refactoring to make use of the updated constructor

src/test/java/org/ethereum/beacon/discovery/NodeRecordTest.java
```diff
@@ -4,7 +4,6 @@
 
 package org.ethereum.beacon.discovery;
 
-import static org.assertj.core.api.Assertions.assertThat;
```
Matilda-Clerke, line 7: Seemed weird to be using multiple assertion APIs, so I removed this one

src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
```diff
@@ -47,36 +47,44 @@ public class NodeRecord {
...
     this.identitySchemaInterpreter = identitySchemaInterpreter;
-  }
-
-  private NodeRecord(IdentitySchemaInterpreter identitySchemaInterpreter, UInt64 seq) {
-    this.seq = seq;
-    this.signature = MutableBytes.create(96);
-    this.identitySchemaInterpreter = identitySchemaInterpreter;
+    this.fields.putAll(fields);
+    // serialise to check size
+    Bytes serializedNodeRecord = serialize();
```
pinges, line 59: Wouldn't it be nice if the NodeRecord was immutable, then we could memoize the serialized record.
It is good that we are checking the resulting length here, but there is not guarantee that the set() method is not called, or the setSignature().
I haven't checked whether we are using the set method, but it would be good to get rid of it!
And what is the business with the signature?
I guess if we don't make the NodeRecord immutable we should check the length when the serialize is called.

src/test/java/org/ethereum/beacon/discovery/NodeRecordTest.java
```diff
@@ -230,23 +231,37 @@ public void testEnrWithEthFieldDecodes() {
...
+    Assertions.assertThrows(
+        IllegalArgumentException.class,
+        () ->
+            NodeRecord.fromValues(
+                IdentitySchemaInterpreter.V4,
+                UInt64.ONE,
+                List.of(
+                    new EnrField(
+                        "test",
+                        Bytes.fromHexString(
```
jframe, line 264: You can use `Bytes.repeat` to make this clearer

src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
```diff
@@ -47,36 +47,46 @@ public class NodeRecord {
...
```
jframe, line 59: I think we should check the size before we RLP decode the node record by checking the bytes size in NodeRecordFactory. It's a shame it can also be directly created so maybe need to keep it on NodeRecord as well?

src/test/java/org/ethereum/beacon/discovery/NodeRecordTest.java
Matilda-Clerke: Will do

src/main/java/org/ethereum/beacon/discovery/schema/NodeRecord.java
Matilda-Clerke: I did look to perform the check in NodeRecordFactory, but the tuweni version in use doesn't give us access to the size of the underlying array. It's possible to get this information by reading the remaining bytes into a Bytes object, taking the size, then creating a new RLPInput to actually read the values out. I thought that might be a bit too messy. What do you think?

jframe: Yeah does sound a bit messy, probably not worth it

jframe (approved):

PR #188, opened 2025-04-15
Add getNodeRecordBuckets, addNodeRecord, and deleteNode to DiscoverySystem

Matilda-Clerke:
Add some methods to aid in the development of the Samba portal client library.

https://github.com/Consensys/discovery/issues/186

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/DiscoverySystem.java
+++ b/src/main/java/org/ethereum/beacon/discovery/DiscoverySystem.java
@@ -131,4 +131,31 @@ public class DiscoverySystem {
         .findFirst()
         .filter(node -> node.getNodeId().equals(nodeId));
   }
+
+  /**
+   * Gets all the KBuckets in the routing table
+   *
+   * @return all the KBuckets in the routing table
+   */
+  public List<List<NodeRecord>> getNodeRecordBuckets() {
+    return buckets.getNodeRecordBuckets();
+  }
+
+  /**
+   * Adds a NodeRecord to the routing table
+   *
+   * @param nodeRecord The NodeRecord to add to the routing table
+   */
+  public void addNodeRecord(NodeRecord nodeRecord) {
+    buckets.offer(nodeRecord);
+  }
+
+  /**
+   * Deletes the NodeRecord identified by nodeId from the routing table
+   *
+   * @param nodeId The node ID to be deleted from the routing table
+   */
+  public void deleteNode(Bytes nodeId) {
+    buckets.deleteNode(nodeId);
+  }
 }
--- a/src/main/java/org/ethereum/beacon/discovery/DiscoverySystem.java
+++ b/src/main/java/org/ethereum/beacon/discovery/DiscoverySystem.java
@@ -62,11 +62,4 @@ public interface DiscoverySystem {
    * @return NodeRecord if any found
    */
   Optional<NodeRecord> lookupNode(final Bytes nodeId);
-
-  /**
-   * Gets all the NodeRecords in the routing table, grouped by their bucket
-   *
-   * @return all the NodeRecords in the routing table, grouped by their bucket
-   */
-  List<List<NodeRecord>> getNodeRecordBuckets();
 }
--- a/src/main/java/org/ethereum/beacon/discovery/DiscoverySystem.java
+++ b/src/main/java/org/ethereum/beacon/discovery/DiscoverySystem.java
@@ -1,59 +1,69 @@
 package org.ethereum.beacon.discovery;
 
-import org.apache.tuweni.bytes.Bytes;
-import org.ethereum.beacon.discovery.schema.NodeRecord;
-import org.ethereum.beacon.discovery.storage.BucketStats;
-
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.stream.Stream;
+import org.apache.tuweni.bytes.Bytes;
+import org.ethereum.beacon.discovery.schema.NodeRecord;
+import org.ethereum.beacon.discovery.storage.BucketStats;
 
 public interface DiscoverySystem {
-    CompletableFuture<Void> start();
-    void stop();
-    NodeRecord getLocalNodeRecord();
-    BucketStats getBucketStats();
-    void updateCustomFieldValue(final String fieldName, final Bytes value);
-    /**
-     * Initiates FINDNODE with node `nodeRecord`
-     *
-     * @param nodeRecord Ethereum Node record
-     * @param distances Distances to search for
-     * @return Future which is fired when reply is received or fails in timeout/not successful
-     *     handshake/bad message exchange. Contains the collection of nodes returned by the peer.
-     */
-    CompletableFuture<Collection<NodeRecord>> findNodes(
-            NodeRecord nodeRecord, List<Integer> distances);
-    /**
-     * Initiates PING with node `nodeRecord`
-     *
-     * @param nodeRecord Ethereum Node record
-     * @return Future which is fired when reply is received or fails in timeout/not successful
-     *     handshake/bad message exchange.
-     */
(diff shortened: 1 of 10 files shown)


Matilda-Clerke:
I have read the CLA Document and I hereby sign the CLA

src/main/java/org/ethereum/beacon/discovery/DiscoverySystem.java
```diff
@@ -131,4 +131,31 @@ public Optional<NodeRecord> lookupNode(final Bytes nodeId) {
...
+  }
+
+  /**
+   * Deletes the NodeRecord identified by nodeId from the routing table
+   *
+   * @param nodeId The node ID to be deleted from the routing table
+   */
+  public void deleteNode(Bytes nodeId) {
+    buckets.deleteNode(nodeId);
+  }
```
garyschulte, line 160: @meldsun0 why does portal client require add/delete node?

garyschulte (commented):
Open question about explicit add/delete.  If we were to run plugin code in the same process space it might pose a risk to allow modifying discovery system.

Also, seems like there is a broken (or flaky) integration test

src/main/java/org/ethereum/beacon/discovery/DiscoverySystem.java
Matilda-Clerke: It seems to be part of the [Portal RPC](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/portal-network-specs/assembled-spec/jsonrpc/openrpc.json) spec to allow direct addition/deletion of nodes. Can we assume that plugin code is trusted, otherwise we wouldn't be running it?

garyschulte: Sorry for the slow response.  This change seems safe presuming that each DiscoverySystem itself would be private to each implementation. (a discv4 instance for besu, and discv5 for samba running as plugin).

@Matilda-Clerke agreed I think we are delegating the decision about plugin code is trusted and safe to the node runner.

garyschulte (approved):

Matilda-Clerke:
Thanks Gary, I'll hold off on merging this until the Teku guys have a chance to take a look too. They're all on leave this week!

src/main/java/org/ethereum/beacon/discovery/DiscoverySystem.java
lucassaldanha, line 160: I think the basic implementation works. But it breaks the assumption that buckets are managed by the underlying discovery system.

I would suggest two possible approaches:
1. Have a flag to enable/disable directly updating buckets in DiscoverySystem (defaulting to false). Then check the flag when doing the operations.
2. Put DiscoverySystem behind an interface and have something like `MutableDiscoverySystem` (or whatever, I am bad with names). Then we only expose these methods add/delete/get in the `MutableDiscoverySystem` interface. Very similar pattern to what we do with WorldState in Besu. Then we could have an option in `DiscoverySystemBuilder` that handles the proper instantiation.

Option 1 is simpler to implement but a bit hacky. I'd suggest using Option 2 as it is more semantic and makes it easier to understand. I would only consider Option 1 if Option 2 ends up being too complicated.

Hopefully either options are not too much of a hassle to implement?

Matilda-Clerke: Definitely prefer option 2. I'll have a look at how it can be done

src/main/java/org/ethereum/beacon/discovery/DiscoverySystemBuilder.java
```diff
@@ -263,7 +263,7 @@ private void createDefaults() {
   LocalNodeRecordStore localNodeRecordStore;
   ExpirationSchedulerFactory expirationSchedulerFactory;
 
-  public DiscoverySystem build() {
+  public <T extends DiscoverySystem> T build(Class<T> desiredClass) {
```
Nashatyrev, line 266: Extracting mutable/immutable interfaces looks like a good thing. 
Just not sure whether we need that complication to return the exact interface here. I would just return `MutableDiscoverySystem` here. 
@lucassaldanha what do you think?

Nashatyrev (commented):
LGTM, just a minor comment

src/main/java/org/ethereum/beacon/discovery/DiscoverySystemBuilder.java
lucassaldanha: Yeah I realised I didn't submit my review comments earlier. My suggestion was to keep the build() method as is. So we don't break the API.

Potentially a separate buildMutable() method or something?

src/main/java/org/ethereum/beacon/discovery/MutableDiscoverySystem.java
```diff
@@ -0,0 +1,23 @@
...
+   * @param nodeRecord The NodeRecord to add to the routing table
+   */
+  void addNodeRecord(NodeRecord nodeRecord);
+
+  /**
+   * Deletes the NodeRecord identified by nodeId from the routing table
+   *
+   * @param nodeId The node ID to be deleted from the routing table
+   */
+  void deleteNode(Bytes nodeId);
```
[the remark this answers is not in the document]
Matilda-Clerke: I don't think it's required for Samba's usecase and it can be added later if desired, so I'll leave it for now.

src/main/java/org/ethereum/beacon/discovery/MutableDiscoverySystem.java
```diff
@@ -0,0 +1,31 @@
...
```
pinges, line 23: nit: naming is inconsistent. Should this be named `deleteNodeRecord` ?

src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java
```diff
@@ -143,4 +144,15 @@ public synchronized Optional<NodeRecord> getNode(final Bytes nodeId) {
   public synchronized boolean containsNode(final Bytes nodeId) {
     return getNode(nodeId).isPresent();
   }
+
+  public synchronized List<List<NodeRecord>> getNodeRecordBuckets() {
```
pinges, line 148: Not sure I like this name. We are not returning the buckets, we are returning node records from all buckets. Maybe a simple name like `getNodeRecords()` is better?

src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java
```diff
@@ -143,4 +144,15 @@ public synchronized Optional<NodeRecord> getNode(final Bytes nodeId) {
   public synchronized boolean containsNode(final Bytes nodeId) {
```
pinges, line 144: I don't understand why these methods are synchronized. E.g. containsNode is always just for a certain point in time. The node record could be deleted just after leaving the synchronized method. Or the node could be added just after we returned false.
Same for the other methods.

pinges (approved):
LGTM, just wanting to understand why these methods need to be synchronized (I know, it wasn't you who started it), and two suggestions for method names.

src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java
Matilda-Clerke: I included the words buckets because each sub list is a list of node records from a bucket. I'm not sure getNodeRecords conveys quite the same meaning.

Personally, I don't understand why node records are organised like that, but it seems important.

src/main/java/org/ethereum/beacon/discovery/storage/KBuckets.java
Matilda-Clerke: I assume it's from an abundance of caution in a system intended to be used in a multi-threaded environment. Perhaps it would be good to overhaul it with ReentractReadWriteLock instead. It could increase performance if the discovery system is used frequently in multiple threads.

PR #191, opened 2025-05-14
Fix concurrency issue on LocalNodeRecordStore

### LINKED ISSUE #190: LocalNodeRecordStore possible concurrency issue when updated it on DiscoveryManagerImpl
Given this method on LocalNodeRecordStore: 

```
    public void onCustomFieldValueChanged(String fieldName, Bytes value) {
        NodeRecord newRecord = oldRecord.withUpdatedCustomField(fieldName, value, this.secretKey);
        this.latestRecord = newRecord;
        this.recordListener.recordUpdated(oldRecord, newRecord);
    }
```

There can be a concurrency issue  even though latestRecord is declared volatile:

```

// ...create newRecord based on oldRecord...
this.latestRecord = newRecord;
```

If two threads call this method at the same time, they may:

Both read the same oldRecord.

Both compute a different newRecord.

Both assign a latestRecord, but the second thread silently overwrites the first one's update.

Suggestions:

- Use synchronized block
- Use AtomicReference<NodeRecord>

lucassaldanha:
## PR Description
- Prevents concurrency issue on LocalNodeRecordStore when checking and comparing the latest record;
- Updated logic to only notify the record listener updates when the record is actually updated (address or custom field);
- Updated unit tests to reflect changes in listener update;

## Fixed Issue(s)
fixes #190

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
+++ b/src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
@@ -5,13 +5,14 @@
 package org.ethereum.beacon.discovery.storage;
 
 import java.net.InetSocketAddress;
+import java.util.concurrent.atomic.AtomicReference;
 import org.apache.tuweni.bytes.Bytes;
 import org.apache.tuweni.crypto.SECP256K1.SecretKey;
 import org.ethereum.beacon.discovery.schema.NodeRecord;
 
 public class LocalNodeRecordStore {
 
-  private volatile NodeRecord latestRecord;
+  private final AtomicReference<NodeRecord> latestRecord;
   private final SecretKey secretKey;
   private final NodeRecordListener recordListener;
   private final NewAddressHandler newAddressHandler;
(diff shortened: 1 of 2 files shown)


src/test/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStoreTest.java
```diff
@@ -42,18 +56,63 @@ public Update(NodeRecord oldRec, NodeRecord newRec) {
...
+    List<Update> listenerCalls = new ArrayList<>();
+    LocalNodeRecordStore recordStore =
+        new LocalNodeRecordStore(
+            nodeRecord, null, (o, n) -> listenerCalls.add(new Update(o, n)), ADDRESS_UPDATER);
+    assertThat(listenerCalls).isEmpty();
+
+    recordStore.onCustomFieldValueChanged("fieldName", Bytes.fromHexString("0x111111"));
+
+    // No updates propagated to listener
+    assertThat(listenerCalls).hasSize(0);
```
rolfyone, line 86: `isEmpty`

rolfyone (approved):
LGTM

src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
```diff
@@ -21,33 +22,43 @@ public LocalNodeRecordStore(
...
     newAddressHandler
-        .newAddress(oldRecord, newAddress)
+        .newAddress(latestRecord.get(), newAddress)
         .ifPresent(
-            record -> {
-              this.latestRecord = record;
-              if (!record.equals(oldRecord)) {
-                recordListener.recordUpdated(oldRecord, record);
+            newRecord -> {
+              final NodeRecord oldRecord = latestRecord.getAndSet(newRecord);
```
Nashatyrev, line 40: Does it still look as non-atomic update here? `get` and `getAndSet` are not under lock and the value can be modified between these calls.

lucassaldanha: Haven't looked into this again. Busy getting PeerDAS into master :)

Will have a second look soon!

PR #192, opened 2025-05-22
Adopt new Tuweni API changes

lu-pinto:
## PR Description
The Tuweni library is being refactored (specifically `org.apache.tuweni.bytes`) to make it more efficient and have a smaller class hierarchy with the aim to make Besu more efficient: https://github.com/Consensys/tuweni/pull/40

Since this project is a dependency of Besu it will need to be aligned with v2 so this PR presents some preparatory work. It is still a WIP as the Tuweni API could change until Besu or other dependents are finalized.


lu-pinto:
I have read the CLA Document and I hereby sign the CLA

gfukushima:
just seen this stale PR, do we have a stable release of tuweni, instead of the snapshot, that we can update and test with?

PR #219, opened 2026-04-22
fix: resend original WHOAREYOU during handshake retransmissions

Gabriel-Trintinalia:
When a session is in `WHOAREYOU_SENT` and another unauthorised packet arrives (e.g. a retransmission), we were generating a new WHOAREYOU with a fresh nonce. 

That breaks the handshake: the initiator has already signed against the original challenge, so the next `HANDSHAKE` gets rejected.

This change reuses the original `WHOAREYOU` instead of creating a new one, keeping the challenge stable and the handshake valid.

 The [spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#session-creation) requires this behaviour:               
   
  ▎ If the recipient has already sent a WHOAREYOU for the node, it should re-send the same WHOAREYOU packet.       

- Fix `HandshakeResend` hive test

---

> [!NOTE]
> **Medium Risk**
> Touches discv5 session handshake state and packet sending logic; incorrect nonce/challenge reuse could break interoperability or leave sessions stuck, though the change is scoped and covered by new tests.
> 
> **Overview**
> Prevents handshake failures on unauthorized packet retransmissions by **re-sending the previously issued `WHOAREYOU`** when the session is already in `WHOAREYOU_SENT` *and* the incoming packet nonce matches (`UnauthorizedMessagePacketHandler`).
> 
> Adds session support for this via `NodeSession` tracking a `pendingWhoAreYouPacket`, exposing `getPendingWhoAreYouNonce()` and `resendOutgoingWhoAreYou()` (reusing the original masking IV/challenge), and clearing the pending packet when handshake state resets. New unit tests cover resend vs. new-challenge behavior and reset/nonce-stability guarantees.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3c39196b2d71d69a54a095df854d7c5db8f4489. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/pipeline/handler/UnauthorizedMessagePacketHandler.java
+++ b/src/main/java/org/ethereum/beacon/discovery/pipeline/handler/UnauthorizedMessagePacketHandler.java
@@ -41,6 +41,15 @@ public class UnauthorizedMessagePacketHandler implements EnvelopeHandler {
                 envelope.getIdString()));
 
     NodeSession session = envelope.get(Field.SESSION);
+
+    // If already awaiting handshake completion, resend the original WHOAREYOU so the
+    // initiator can complete it using the same challenge nonce, rather than issuing a
+    // new challenge with the retransmitted packet's nonce.
+    if (session.getState() == SessionState.WHOAREYOU_SENT) {
+      session.resendOutgoingWhoAreYou();
+      return;
+    }
+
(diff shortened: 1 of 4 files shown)


macfarla (approved):
LGTM

PR #202, opened 2026-03-20
Fix ipv6 discovery

gfukushima:
## PR Description
Vibe coded fix for the external address selection "bug" that was preventing teku from obtaining the ipv6 from discovery. 

## Fixed Issue(s)

Part of [#10305](https://github.com/Consensys/teku/issues/10305)

---

> [!NOTE]
> **Medium Risk**
> Changes external address auto-selection to update IPv4 and IPv6 ENR fields independently, which can affect how nodes advertise reachable endpoints. Main risk is incorrect ENR updates (e.g., flapping or missing one family) impacting peer connectivity in dual-stack environments.
> 
> **Overview**
> Fixes dual-stack external address auto-discovery by selecting the best *IPv4* and *IPv6* external addresses independently instead of letting one IP family dominate and prevent the other from being published.
> 
> `DefaultExternalAddressSelector` now runs separate vote-based selection for IPv4/IPv6 and updates the corresponding ENR field via a shared `maybeUpdateAddress` path.
> 
> Test utilities and coverage are extended to support IPv6 ENR fields: `SimpleIdentitySchemaInterpreter` can now create/read `udp6` records and preserves the other family when updating an address, and new `DefaultExternalAddressSelectorTest` cases validate both families can be discovered and that IPv6 selection is independent of IPv4 vote volume.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7db722859fbc17a327ee08050deac48fc3bfc988. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #225, opened 2026-05-11
Fix IPv6 outbound dial bootstrap

usmansaleem:
## Summary

When a node binds an IPv6 discovery socket but its home record has not (yet) advertised `ip6`, `NodeSessionManager.getRemoteSocketAddress` would still pick the peer's IPv4 address for outbound dials. This makes IPv6 external-address auto-discovery impossible to bootstrap: peers can only PONG with our IPv6 source address if we PING them over IPv6 first, but we never do.

### Current logic

```java
if (homeUdpAddress.isPresent() && homeUdp6Address.isPresent()) {
    return nodeRecord.getUdp6Address().or(nodeRecord::getUdpAddress);
} else if (homeUdp6Address.isPresent()) {
    return nodeRecord.getUdp6Address();
} else {
    return nodeRecord.getUdpAddress();   // ← IPv4-only home record falls here
}
```

A dual-stack-bound node whose home record advertises only `ip` always dials IPv4 → peers PONG with our IPv4 recipientIp → `DefaultExternalAddressSelector` accumulates only IPv4 votes → MIN_CONFIRMATIONS=2 for IPv6 is never reached → the `NewAddressHandler` never fires for IPv6 → our home record never gains `ip6`. Chicken-and-egg.

### Fix

Thread a `boolean ipv6BindAvailable` from `DiscoveryManagerImpl` (derived from the bound `discoveryServers` list) into `NodeSessionManager`. When set, prefer the peer's `ip6` — even if our home record has only `ip` — falling back to the peer's `ip` if the peer is IPv4-only.

```java
final boolean preferIpv6Outbound = homeUdp6Address.isPresent() || ipv6BindAvailable;
if (homeUdpAddress.isPresent() && preferIpv6Outbound) {
    return nodeRecord.getUdp6Address().or(nodeRecord::getUdpAddress);
} else if (preferIpv6Outbound) {
    return nodeRecord.getUdp6Address();
} else {
    return nodeRecord.getUdpAddress();
}
```

Together with the per-family selector change in #202 (already merged on master) this lets a dual-stack-bound node populate its own `ip6`/`udp6`/`tcp6` ENR fields via peer consensus, fixing the auto-discovery bootstrap that downstream feature [besu-eth/besu#9874](https://github.com/besu-eth/besu/issues/9874) depends on.

### Why this hasn't surfaced before

The discovery library currently has two consumers, Teku and Besu, and they populate the home node's ENR with different timing:

- **Teku** populates the local ENR's `ip6` field at startup, either from `--p2p-advertised-ip` or from local-interface resolution (`NetworkConfig.getLocalAddress(IP_V6)`), before discovery starts. So `homeUdp6Address.isPresent()` is true whenever Teku has an IPv6 socket bound — the existing v6-preference path is already taken, and Teku has never observed the bug. See Teku commit [`bf348b7271`](https://github.com/Consensys/teku/commit/bf348b7271) ("Fix ipv6 discovery — always populate the initial ENR with the resolved listen addresses").
- **Besu**'s in-flight peer-consensus IPv6 auto-discovery ([besu-eth/besu#10416](https://github.com/besu-eth/besu/pull/10416), for [besu-eth/besu#9874](https://github.com/besu-eth/besu/issues/9874)) intentionally leaves `ip6` empty at boot and relies on the external-address selector to populate it from peer PONGs. This places the home record in exactly the state the current dialer refuses to handle: socket bound, ENR empty.

This PR is therefore a **no-op for Teku** — when the home record already advertises `ip6`, `homeUdp6Address.isPresent()` is true and dominates the new OR, so dial selection is byte-for-byte identical to today. For Besu it closes the bootstrap loop together with the per-family selector change already on master in #202.

### Verified

- Compiles clean.
- Local Docker end-to-end against a Besu build pinned to this snapshot: after ~75s and 2 connected peers, the target node's ENR transitions from IPv4-only `seq=1` to dual-stack `seq=2` with the correct `ip6`/`udp6`/`tcp6` values. Without this fix, `seq` stays at 1 with no IPv6 fields.
- Added two unit tests in `NodeSessionManagerTest`:
  - `shouldDialPeerOverIpv6WhenIpv6BindAvailableEvenIfHomeRecordIsIpv4Only` — the bootstrap case.
  - `shouldDialPeerOverIpv4WhenIpv6BindNotAvailableAndHomeRecordIsIpv4Only` — regression of the existing default behaviour.

### Behaviour matrix

| Home record | `ipv6BindAvailable` | Peer ENR | Resulting dial |
|---|---|---|---|
| IPv4 only | false (today) | dual-stack | peer IPv4 (unchanged) |
| IPv4 only | **true (new)** | dual-stack | peer IPv6 |
| IPv4 only | **true (new)** | IPv4-only | peer IPv4 (graceful fallback) |
| dual-stack | any | dual-stack | peer IPv6 (unchanged) |
| IPv6 only | any | dual-stack | peer IPv6 (unchanged) |

## Test plan

- [x] `./gradlew compileJava compileTestJava` passes
- [x] New unit tests cover both new and regression behaviours
- [x] Verified end-to-end with a Besu build pinned to this snapshot
- [x] CI on this PR

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/DiscoveryManagerImpl.java
+++ b/src/main/java/org/ethereum/beacon/discovery/DiscoveryManagerImpl.java
@@ -9,6 +9,7 @@ import static org.ethereum.beacon.discovery.util.Utils.RECOVERABLE_ERRORS_PREDIC
 import com.google.common.annotations.VisibleForTesting;
 import io.netty.channel.socket.InternetProtocolFamily;
 import io.netty.channel.socket.nio.NioDatagramChannel;
+import java.net.Inet6Address;
 import java.net.InetSocketAddress;
 import java.util.Collection;
 import java.util.List;
@@ -92,13 +93,22 @@ public class DiscoveryManagerImpl implements DiscoveryManager {
     final NodeRecord homeNodeRecord = localNodeRecordStore.getLocalNodeRecord();
 
     this.discoveryServers = discoveryServers;
+    // Detect whether any of the bound discovery servers is listening on IPv6. The session manager
+    // uses this to allow IPv6 outbound dialing when the home record has not (yet) advertised IPv6
+    // — required for IPv6 external-address auto-discovery to bootstrap.
+    final boolean ipv6BindAvailable =
+        discoveryServers.stream()
+            .map(NettyDiscoveryServer::getListenAddress)
+            .map(InetSocketAddress::getAddress)
+            .anyMatch(addr -> addr instanceof Inet6Address);
     nodeSessionManager =
         new NodeSessionManager(
             localNodeRecordStore,
             signer,
             nodeBucketStorage,
             outgoingPipeline,
-            expirationSchedulerFactory);
+            expirationSchedulerFactory,
+            ipv6BindAvailable);
     incomingPipeline
         .addHandler(new PacketSourceFilter(addressAccessPolicy))
         .addHandler(new IncomingDataPacker(homeNodeRecord.getNodeId()))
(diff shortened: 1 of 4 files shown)


Gabriel-Trintinalia (approved):
lgtm

gfukushima (approved):

PR #229, opened 2026-05-13
Fix concurrent local ENR custom field updates

### LINKED ISSUE #190: LocalNodeRecordStore possible concurrency issue when updated it on DiscoveryManagerImpl

lucassaldanha:
## Summary
- serialize `LocalNodeRecordStore` writer paths so local ENR updates are applied from the latest record and listener callbacks remain ordered
- keep writer methods synchronized while making `latestRecord` volatile so `getLocalNodeRecord()` can read without taking the lock
- expand concurrency regression coverage for custom-field updates, bound-port resolution, and mixed socket-address/custom-field updates
- replace raw thread joins with timeout-based executor helpers and assert listener update chains

Fixes #190

## Tests
- `./gradlew spotlessApply test --tests org.ethereum.beacon.discovery.storage.LocalNodeRecordStoreTest`

### DIFF
--- a/src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
+++ b/src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
@@ -92,9 +92,12 @@ public class LocalNodeRecordStore {
   }
 
   public void onCustomFieldValueChanged(final String fieldName, final Bytes value) {
-    NodeRecord oldRecord = this.latestRecord.get();
-    NodeRecord newRecord = oldRecord.withUpdatedCustomField(fieldName, value, signer);
-    this.latestRecord.set(newRecord);
+    NodeRecord oldRecord;
+    NodeRecord newRecord;
+    do {
+      oldRecord = this.latestRecord.get();
+      newRecord = oldRecord.withUpdatedCustomField(fieldName, value, signer);
+    } while (!latestRecord.compareAndSet(oldRecord, newRecord));
     recordListener.recordUpdated(oldRecord, newRecord);
   }
 }
--- a/src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
+++ b/src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
@@ -7,14 +7,13 @@ package org.ethereum.beacon.discovery.storage;
 import java.net.Inet6Address;
 import java.net.InetSocketAddress;
 import java.util.Optional;
-import java.util.concurrent.atomic.AtomicReference;
 import org.apache.tuweni.bytes.Bytes;
 import org.ethereum.beacon.discovery.crypto.Signer;
 import org.ethereum.beacon.discovery.schema.NodeRecord;
 
 public class LocalNodeRecordStore {
 
-  private final AtomicReference<NodeRecord> latestRecord;
+  private NodeRecord latestRecord;
   private final Signer signer;
   private final NodeRecordListener recordListener;
   private final NewAddressHandler newAddressHandler;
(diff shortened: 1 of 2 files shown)


src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
usmansaleem: @lucassaldanha You probably need to apply same structure here.

lucassaldanha:
@usmansaleem wnded up switchin from the CAS loop to synchronized access because `LocalNodeRecordStore` has multiple writer paths, not just custom-field updates. A CAS loop in one method fixes only that method’s lost-update case, while `onSocketAddressChanged`, `onBoundPortResolved`, and custom-field writes can still race with each other.

  Serializing the writer methods treats the local ENR as one shared state machine: each update is computed from the latest record, applied atomically with respect to other writers, and its listener callback is emitted in
  the same order as the state transition. These updates are infrequent, so the simplicity and stronger ordering guarantees are worth the negligible lock cost. And given this isn't a hot path for performance the tradeoff seems acceptable.

src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
```diff
@@ -7,14 +7,13 @@
...
 import java.util.Optional;
-import java.util.concurrent.atomic.AtomicReference;
 import org.apache.tuweni.bytes.Bytes;
 import org.ethereum.beacon.discovery.crypto.Signer;
 import org.ethereum.beacon.discovery.schema.NodeRecord;
 
 public class LocalNodeRecordStore {
 
-  private final AtomicReference<NodeRecord> latestRecord;
+  private NodeRecord latestRecord;
```
usmansaleem: You can add`volatile` modifier here. This would allow us to remove `synchronized` from `public synchronized NodeRecord getLocalNodeRecord() {` at line 32. The writers to latestRecord can remain `synchronized` though.

src/main/java/org/ethereum/beacon/discovery/storage/LocalNodeRecordStore.java
```diff
@@ -58,43 +51,45 @@ record -> {
...
-    this.latestRecord.set(newRecord);
-    recordListener.recordUpdated(oldRecord, newRecord);
+  public synchronized void onCustomFieldValueChanged(final String fieldName, final Bytes value) {
+    final NodeRecord oldRecord = this.latestRecord;
+    final NodeRecord newRecord = oldRecord.withUpdatedCustomField(fieldName, value, signer);
+    updateLocalNodeRecord(oldRecord, newRecord, true);
+  }
+
+  private void updateLocalNodeRecord(
+      final NodeRecord oldRecord, final NodeRecord newRecord, final boolean notifyWhenUnchanged) {
```
usmansaleem: nit: `notifyWhenUnchanged` probably doesn't matter, if `notifyWhenUnchanged` is true (called from onCustomFieldValueChanged), the new record will never match old record (seq would have bumped). Not a show-stopper though.

usmansaleem (approved):
LGTM

PR #238, opened 2026-08-10
Feature: Remove vertx as a dependency / transitive deps via tuweni

garyschulte:
## PR Description

remove vertx as an explicit dependency and instead rely on transitive deps from tuweni.  This prevents pulling in multiple versions of vertx into discovery, tuweni, besu.

vertx has not been a direct dependency since #139, four years ago

Ideally this PR will wait to merge until tuweni is updated to the latest vertx in [65](https://github.com/Consensys/tuweni/pull/65) and version 2.8.0 released.  

## Fixed Issue(s)

---

> [!NOTE]
> **Medium Risk**
> Bumps tuweni (bytes/crypto/rlp), which can change transitive Vert.x resolution across consumers. The session cleanup tweak is functionally equivalent for `NodeSession`.
> 
> **Overview**
> Removes the unused direct `vertx-core` dependency and bumps **tuweni** from `2.7.0` to `2.8.0`, so Vert.x comes in only transitively and version conflicts with tuweni/Besu are avoided.
> 
> Also switches session cleanup in `NodeSessionManager` from reference equality (`==`) to `.equals()` when removing nonce mappings for a deleted session.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 42bb9be63654d0a2841a52ed3942e832fa8d0619. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #239, opened 2026-08-12
Revert to identity comparison

garyschulte:
## PR Description
Revert value comparison change in #238  in NodeSessionManager cleanup. 

In this case `.equals()` and `==` are the same, but using the identity comparison better conveys intent and is more future-proof if NodeSession overrides `.equals()` in the future

## Fixed Issue(s)

---

> [!NOTE]
> **Low Risk**
> Single-line behavioral change in session teardown; with default `Object` equality it matches prior `.equals` behavior but documents intent for future `NodeSession` equality changes.
> 
> **Overview**
> In **`NodeSessionManager.deleteSession`**, nonce cleanup in `lastNonceToSession` again uses **reference equality** (`s == removedSession`) instead of `.equals(removedSession)`, reverting the change from #238.
> 
> Comments clarify that only mappings for the **exact session instance** being torn down should be removed, so a newer session object must not be cleared if it were ever considered equal by value.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 29001f96d44eb7ac0447286f426afddd37bea03e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


siladu (approved):

usmansaleem (approved):
LGTM

